PR #127, opened 2015-01-08
Updates copyright to 2015

Uladox:
Uses the grep -rl '2014' ./ | xargs sed -i 's/2012/2015/g' command in the src directory

### DIFF
--- a/src/reader_struct.h
+++ b/src/reader_struct.h
@@ -269,31 +269,6 @@ struct TypeReader<T, Category::Primitive> {
 };
 
 /**
- * Structure field reader.
- */
-template <class S, class T>
-struct FieldReader {
-	static void ReadLcf(S& obj, T S::*ref, LcfReader& stream, uint32_t length) {
-		TypeReader<T>::ReadLcf(obj.*ref, stream, length);
-	}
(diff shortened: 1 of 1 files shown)


src/generated/rpg_eventcommand.h
```diff
@@ -126,8 +126,8 @@ namespace RPG {
 				ConditionalBranch_B = 13310,
 				TerminateBattle = 13410,
 				ShowMessage_2 = 20110,
-				ShowChoiceOption = 20140,
-				ShowChoiceEnd = 20141,
+				ShowChoiceOption = 20150,
+				ShowChoiceEnd = 20151,
```
Ghabry: .

Uladox: Oh, I better fix that.

Ghabry:
I don't really see a reason in updating copyrights. We just putted the first copyright year there (when the lib switched to MIT license). And copyright lasts for at least 70 years.
And you forgot updating the templates in the template directory. And then just run the generate.py script.

Uladox:
Go point, got caught up in seeing something easy to update and reacted prematurely.

PR #131, opened 2015-02-27
Use AX_PKG_CHECK_MODULES to set library dependencies (fixes #129)

### LINKED ISSUE #129: Hardcoded library dependencies on liblcf.pc.in
This affects at least autotools method, where ./configure allows to disable expat (disable XML support) and disable ICU (and fallback to iconv). When Player or any other app is using pkg-config to check for liblcf it will fail because other ICU dependencies may not be available according with installed liblcf.pc.

fdelapena:


fdelapena:
Jenkins: retest this please

PR #132, opened 2015-03-11
Identify a lot of fields and fix a bug when loading actor sprites

Zegeri:
This will conflict with the current EasyRPG Player, because SavePartyLocation.original_move_route_index has been renamed to SavePartyLocation.boarding.

If you have an event with the commands
-Key Input Proc
-Open save menu
Those SaveEvents.keyinput_something fields will be activated. If the event was a parallel process, that information will be in SaveEventData.keyinput_something fields. However, I still don't know what is all that for. Redundant data? Or maybe it tells us more about the inner workings of the original interpreter?


generator/csv/fields.csv
```diff
@@ -909,7 +909,8 @@ SaveVehicleLocation,flash_current_level,f,Double,0x54,0.0,double
 SaveVehicleLocation,flash_time_left,f,Integer,0x55,0,int
 SaveVehicleLocation,unknown_65,f,Integer,0x65,0,?
 SaveVehicleLocation,original_move_route_index,f,Integer,0x66,0,Index of custom move route
-SaveVehicleLocation,unknown_6a,f,Integer,0x6A,0,?
+SaveVehicleLocation,remaining_ascent,f,Integer,0x6A,0,From 0 to 255 - In flying vehicles, remaining distance to ascend
+SaveVehicleLocation,remaining_descent,f,Integer,0x6B,0,From 0 to 255 - In flying vehicles, remaining distance to descend
```
fdelapena: Lines 912 and 913 have extra "," in the comment column and breaks GitHub parser for web visualization. They could use ";" instead.

Zegeri:
I'm going to add 2 more unknown chunks. You can find System.unknown_61 in the RPG_RT.ldb of El Gofre Misterioso and SaveEventData.unknown_04 when leaving unfinished an event called from a parallel common event (I hope it'll make some sense soon).

PR #139, opened 2015-05-10
dst_enc wasn't converted to the proper names ICU/iconv expect

Ghabry:
No idea if this also fixes saving in emscripten... but is possible

Should fix saving when the encoding was specified in the INI (why was this never noticed) (fixes EasyRPG/Player#465).

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -240,13 +240,17 @@ std::string ReaderUtil::Recode(const std::string& str_to_encode, const std::stri
 std::string ReaderUtil::Recode(const std::string& str_to_encode,
                                const std::string& src_enc,
                                const std::string& dst_enc) {
-	std::string encoding_str = src_enc;
+	std::string src_enc_str = src_enc;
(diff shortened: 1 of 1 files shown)


fdelapena:
It does not work for me with Don's adventures with existing Encoding=1251, after building and installing ghabry-liblcf:development and rebuilding a clean player linked with it they are being saved corrupted yet.

Ghabry:
Saving was totally broken when passed via "--encoding", because it will just use the original value, e.g. "1252" which is invalid to ICU/iconv.
This got fixed by passing dst_enc to CodepageToEncoding.
The same bug wasn't visible for the ini-case, because GetEncoding called CodepageToEncoding already. So saving worked for encoding that have a specific hardcoded mapping, like Japanese, but Russian just got windows-1252, which is ambigious. Ambigious was already accepted for the src, but not for the dst case.
To sum up: When you had a russian game and it was NOT autodetected, then saving failed.

PR #153, opened 2015-09-15
Add new RPG2k 1.61 and RPG2k3 1.11 fields

Ghabry:
version is currently 2-state (0 or 1) but maybe this gets more numbers later, so made this Integer to be future proof 

At least we have sourcecode, making this straight-forward


fdelapena:
Build finished. No test results found.

Ghabry:
The white space is significant then in Python :D

carstene1ns:
Just autotools being picky. Our fault for treating warnings as errors.

fdelapena:
@Ghabry neat, could you move the clcompile line after read_struct? ;)

There is a new checkbox in 2k3 1.11 about some save field, is this included here?

Ghabry:
I can't find that field, can I have a screenshot from a person who found it?

fdelapena:
It was not a new field, sorry. Just the layout was different.

src/generated/ldb_chunks.h
```diff
@@ -1276,7 +1276,9 @@ namespace LDB_Reader {
 			/** String - RPG2003 */
 			frame_name				= 0x64,
 			/** Flag - RPG2003 */
-			invert_animations		= 0x65 
+			invert_animations		= 0x65,
+			/** When false the title is skipped and the game starts directly. In TestPlay mode skips directly to the Load scene. Added in RPG Maker 2003 v1.11 */
```
CherryDT: This is not entirely true - it starts the load scene only when HideTitle is used (not just test play).

PR #160, opened 2015-11-18
Change Korean and Japanese encodings to fix file finding errors

fdelapena:
Fixes EasyRPG/Player#649

ICU from toolchains will need to be rebuilt to make this update work properly.

After updating toolchains also closes EasyRPG/Player#746

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -47,7 +47,7 @@ std::string ReaderUtil::CodepageToEncoding(int codepage) {
 
 	if (codepage == 932) {
 #ifdef LCF_SUPPORT_ICU
-		return "ibm-943_P130-1999";
+		return "ibm-943_P15A-2003";
 #else
 		return "SHIFT_JIS";
 #endif
@@ -134,7 +134,7 @@ std::string ReaderUtil::DetectEncoding(const std::string& database_file) {
 		// Fixes to ensure proper Windows encodings
 		if (encoding == "Shift_JIS")
 		{
-			encoding = "ibm-943_P130-1999"; // Japanese with Yen as backslash
+			encoding = "ibm-943_P15A-2003"; // Japanese with \ as backslash
 		}
 		else if (encoding == "EUC-KR")
 		{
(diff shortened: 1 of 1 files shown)


Ghabry:
Is that encoding not supported in the static toolchains? :(

fdelapena:
Not tested, the icudata demo should explain codepage dependencies better. I can check it tonight.

carstene1ns:
datfile

Ghabry:
You can now embed word documents to github comments iirc. So you can embed "dat file" in the word document and upload that one :D

fdelapena:
Here is the [.dat download](https://fran.cr/owncloud/index.php/s/Z4FqmlqS2z2iiRX/download) link, for ICU56

Ghabry:
[datfile.docx](https://github.com/EasyRPG/liblcf/files/39229/datfile.docx)

fdelapena:
After further testing, yet another issue. This time with middle dot conversion into UTF-8:

When trying to read some file with middle dot, we have 2 possibilities, depending on encoding:
U+00B7 · middle dot (windows-949-2000, uses backslash)
U+30FB ・ katakana middle dot (ibm-1363_P110-1997, uses Won)

Test: `--encoding windows-949-2000 --start-map-id 11 --new-game`

Not sure which filename is the right on real win32 RTP, as online decoders might have differ from real Win9x. At least after using convmv to unpack the zip, it needs windows-949-2000 to load this file properly.

carstene1ns:
time to get a windows 98 vm ready :grinning:

fdelapena:
Can't test this until Player#673 is fixed.

fdelapena:
After unpacking a zip using Korean locale on Windows, it seems to use U+00B7, so windows-949-2000 (with \ ) is the current candidate.

Ghabry:
Is this still pending fix?

fdelapena:
> Is this still pending fix?

Fixed, tested and ready.

fdelapena:
Don't merge yet, it looks like Japanese encoding will need an update too... it does not detect filanemes with ～ properly and the new encoding kills backslash to yen conversion :sob:.

fdelapena:
Jenkins: retest this please

fdelapena:
Ready for your ICU rebuilds. ^^

PR #161, opened 2015-12-06
Clear all database data in Data::Clear

Ghabry:
Required by the Game Browser to get rid of System changes of the previous launched game


Ghabry:
Just for informaton: This is not related to any F12-issues, this issues fixes problems with the ldb database, not with the editable RPG::Save data.

PR #165, opened 2016-01-21
Correctly calculate size of RPG::EventCommand

Ghabry:
This should Fix https://github.com/EasyRPG/Player/issues/556

Can anybody else confirm this? Delete your old savegames, if they are already corrupted it won't help :D

Please also test: Save a game in the house with 1252 encoding. Then load it with 932 and save it.
It should still work (ICU substitutes invalid codepoints silently), at least for me.


fdelapena:
Cross-tested with reading and saving with both encodings, it works flawlessly, not crashing anymore :+1:.

PR #180, opened 2016-05-10
SaveActor 0x60 -> battler_animation; Rename Class->unarmed_animation to battler_animation

Ghabry:
Basicly the battle animation of the actor... no idea why this exists but RPG 2k3 reads from it ...
Also fixes bad naming in the Class-class. Does not break Building because we dont use it :)

Related https://github.com/EasyRPG/Player/issues/876


Ghabry:
Please don't merge yet, have to figure out the default value first :)

PR #181, opened 2016-05-23
Use strings from Data::system to detect encoding

Zegeri:
This improves the encoding detection of translated games that use the RTP of the original language.

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -76,7 +76,7 @@ std::string ReaderUtil::DetectEncoding(const std::string& database_file) {
 #ifdef LCF_SUPPORT_ICU
 	std::ostringstream text;
 
-	//Populate Data::terms or will empty by default even if load fails
+	//Populate Data::terms and Data::system or will empty by default even if load fails
 	LDB_Reader::Load(database_file, "");
 
 	text <<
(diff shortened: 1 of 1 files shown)


Ghabry:
I wonder btw if ICU really analyzes the whole string or only a few bytes.

Zegeri:
It probably analyzes the string until one of the encoding candidates reaches 100 confidence. Let's see how does the detector act after this changes with the english fan translation of Yume Nikki.
Before:
```
Input: Quit Start Diary Quit           E $k            Choose a page to write Choose a page to read Page End game and return to menu? Yes No
Encoding(Confidence): ISO-8859-1(61),ISO-8859-2(37),UTF-8(15),UTF-16BE(10),UTF-16LE(10),Shift_JIS(10),GB18030(10),EUC-JP(10),EUC-KR(10),Big5(10),ISO-8859-9(2
```
After:
```
Input: Quit Start Diary Quit           E $k            Choose a page to write Choose a page to read Page End game and return to menu? Yes No vehicle vehicle vehicle �^�C�g�� gameover �}�C�V�X�e��a system2a ���� 
Encoding(Confidence): Shift_JIS(100),windows-1252(49),windows-1250(40),windows-1254(14),UTF-16BE(10),UTF-16LE(10),GB18030(10),Big5(10)
```

fdelapena:
By the way, some time ago I was reconsidering that space concatenation removal because it looks like it is character based only, I didn't test if this affects the confidence value. At least Japanese does not use too much spaces.

fdelapena:
This will fix multiple "file not found" issues with lots of Japanese games, thanks!

Games using broken mojibake in the filesystem are a minority fortunately, these users might need to  adjust the encoding by hand (mostly bad game translations).

carstene1ns:
btw. I also had in mind to make the detection string longer :grinning:.
https://github.com/carstene1ns/easyrpg-tools/blob/encdetect/encdetect/src/main.cpp#L187-202
About confidence value, I think we can ignore everything lower than a certain value (i.e. 20).

Ghabry:
Got the first report on google play today about a Spanish player playing Mojibake-YumeNikki. Before this was detected as European, now is detected as Japanese because of the filenames. But is a minor problem as fdelepena said.

PR #184, opened 2016-08-10
Fix up SaveMapInfo.chipset_id value when equal or less than 0

Zegeri:
This fixes a crash when loading some save files since EasyRPG/Player#979.


Ghabry:
Some users reported crashes when loading saves. Is probably this one.

PR #192, opened 2016-10-17
Add C++11 to time_stamp test

fdelapena:


Ghabry:
The lines look fine but delegating this to @carstene1ns because I have no idea if this is the best way to do it or to place it.

carstene1ns:
Yeah, looks good to me.

Some history: wanted to initially solve this with [AX_CXX_COMPILE_STDCXX(11)](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html), because gcc < 4.8 does not support `-std=c++11` and gcc > 6 does not need this flag anymore, so it would be seamlessly enabled, if needed.

PR #199, opened 2017-04-12
Fix savegame corruption introduced in 5abb593d

Ghabry:
Removed the asserts and changed the error handling a bit because the broken event commands are reallllly bad and result in an assert when opening Scene_load or Scene_Save otherwise :(

Fixes EasyRPG/Player#1149.


Ghabry:
Yepp I just noticed that I didn't even revert the broken line >.<

Ghabry:
We are actually lucky that the event data is the last in the savefile, so loading such a corrupted save fixes itself in most cases by simply reentering the map 👍  (not possible with Yume Nikki, stuck at "Loading: Please Wait" SE preload).

Tested with VH1, Yume2kki and Yume Nikki: The save/load scene opens without crashing and after loading the broken save and resaving the errors are gone.

PR #202, opened 2017-06-11
Replace code generator with Jinja2 templates

Ghabry:
This is the first step for "liblcf vNext". I need a more maintanable code generator :).
generate.py depends now on the template engine "Jinja2".
Not tested under Python 2 yet. Need to boot my other system first.

The result matches the old generator output except for:
- Whitespace in *_chunks.h and *_flags.cpp. Too much work to implement the tab-insertion logic and these files are expected to be removed in the next refactor.
- Whitespace between functions and fields in *.h. Difficult to add because unconditionally adding a whitespace is not possible because this adds an incorrect whitespace after "public:" when no functions exist.


carstene1ns:
Tbh. I did not really like this excessive whitespace usage much.

Ghabry:
oh just noticed that the branch already contains more changes like Get/Set functions that I don't want in the PR... meh, making a total mess... :/

Ghabry:
Can confirm that the script works with 0 zeros in python2 👍  (lucky)

carstene1ns:
much easier to review: https://github.com/EasyRPG/liblcf/pull/202/files?w=1 😀

Ghabry:
Found a problem: When building with CMake the build fails because when Python is installed it uses the file listing feature to get the list of generated files for compiling and the listing feature is broken. WTF.

I'm considering to delete the "-l" feature because I find it quite useless and fix the CMake script.

Any objections?

Ghabry:
"Fixed" the CMake script by removing special path when Python is installed. Was less work then fixing the script, sorry :)

PR #206, opened 2017-08-03
Provide i/o stream API for liblcf

Ghabry:
Finishes the work of ChrisBreiti and provides a compatibility layer around the old function making this change backward compatible 👍 

The code is required for the WIP filesystem abstraction in Player. FILE* can't be used here because it can only operate on "real" files and we can't use the non-standard fopencookie. Soooooo C++ inheritance to the rescue.


src/writer_xml.h
```diff
@@ -135,4 +122,4 @@ class XmlWriter {
 
 };
 
-#endif
+#endif
```
carstene1ns: \n

Ghabry: Done, thx

Ghabry:
This PR crashes under Android with:
```
********** Crash dump: **********
Build fingerprint: 'Huawei/U8860/hwu8860:4.0.3/HuaweiU8860/CHNC00B932SP01:user/ota-rel-keys,release-keys'
pid: 1766, tid: 1881  >>> org.easyrpg.player <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 00000000
Stack frame #00  pc 001ddcb8  /data/data/org.easyrpg.player/lib/libmain.so: Routine __exchange_and_add at /tmp/ndk-user/tmp/gnu-libstdc++/static-armeabi-v7athumb-4.8/include/ext/atomicity.h:49
Stack frame #01  pc 001da3d2  /data/data/org.easyrpg.player/lib/libmain.so: Routine std::__timepunct<char>::_M_put(char*, unsigned int, char const*, tm const*) const at /tmp/ndk-user/tmp/gnu-libstdc++/static-armeabi-v7athumb-4.8/src/c++98/time_members.cc:60
Stack frame #02  pc 001c6c26  /data/data/org.easyrpg.player/lib/libmain.so: Routine std::basic_ios<wchar_t, std::char_traits<wchar_t> >::_M_cache_locale(std::locale const&) at /tmp/ndk-user/tmp/gnu-libstdc++/static-armeabi-v7athumb-4.8/include/bits/basic_ios.tcc:170
```
So our Android build is currently broken.

android-rage-out

PR #213, opened 2017-09-18
Add Rpg2k3 1.12 Chunks

Ghabry:
Only 90% sure that the Layer enums are correct, though.

Ping @20kdc


carstene1ns:
I do not think we should change CSV format.

Ghabry:
Okay, how shall the comma in the comment column be handled?

Imho it would be good to add proper comments to fields.csv instead of "Integer" or "Boolean", thats why I proposed it here.

Ghabry:
Deleted the first commit with the ","-parser change

PR #216, opened 2017-10-11
Use `doctest` for unit tests

carstene1ns:
This disables the autotools (parallel) and cmake (ctest) test drivers and uses the executable directly.

Not quite happy with the CMake change. The `test` target is reserved, so I used `check` like autotools.
This is done by [CMP0037](https://cmake.org/cmake/help/v3.0/policy/CMP0037.html) which was introduced in 3.0.
We currently target 2.8, which means we would need to conditionally enable it like this to stay compatible:
```patch
+  cmake_policy(PUSH)
+  if(POLICY CMP0037)
+    cmake_policy(SET CMP0037 OLD)
+  endif()
   add_custom_target(test)
+  cmake_policy(POP)
```
So question is:
- use `check` target
- stay with `test` target
  - stay with ctest (unneeded, does not display test output)
  - bump to cmake 3.0 and enable the policy


Ghabry:
ctest is really inconvenient, even with --verbose it still eats the color.

I'm fine with using "make check" for both build systems.

Ghabry:
Used "cmake --build . --target check -- -j4" via Makefile and Ninja.
Works for both, though Ninja eats the colors :(

carstene1ns:
Unfortunately there is no easy way with the custom target to pass command line parameters from the environment to the test_runner :/

PR #220, opened 2017-12-13
Fixes for GCC 4.8.x

gameblabla:
These small changes (just adding #include <stdio.h> to 2 files) are required for GCC 4.8.x, which happens to be the default GCC version for the GCW0. (Yes, they're still stuck with that old version)

Without this pr, i get undefined references to fprintf.

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -31,6 +31,7 @@
 #   include <locale>
 #endif
 
+#include <cstdio>
 #include <cstdlib>
 #include <sstream>
 #include <vector>


Ghabry:
Please use <cstdio> instead of <stdio.h> because this is C++ code (just for style, has no relevant semantic change)

src/reader_lcf.cpp
```diff
@@ -7,6 +7,7 @@
  * file that was distributed with this source code.
  */
 
+#include <stdio.h>
```
Ghabry: This file already has this include for the NDEBUG case a few lines below. Please remove the 3 lines of NDEBUG check below, too.

gameblabla:
Alright, this should do it now. It compiles properly.

src/reader_lcf.cpp
```diff
@@ -7,6 +7,7 @@
  * file that was distributed with this source code.
  */
 
+#include <cstdio>
```
Ghabry: Could you remove the
```cpp
#ifdef NDEBUG
#  include <stdio.h>
#endif
```
that is a few lines below? Is not needed anymore after adding this.

gameblabla: Well, NDEBUG wasn't defined by default for the GCW0 build so i removed that line (as you requested) and added NDEBUG to OpenDingux's makefile.

PR #222, opened 2018-01-01
Change all int types to int32_t.

Ghabry:
Additionally some cases of "> >" were changed to >> and reader_types.h was replaced with stdint.h.


src/lcf_options.h
```diff
@@ -21,6 +21,6 @@
 /**
  * Outputs information about read chunks on stdout during parsing
  */
-//#define LCF_DEBUG_TRACE
```
carstene1ns: debug leftover?

Ghabry: yes :/

PR #223, opened 2018-02-08
Fix incorrect string handling of istream ini reader

Ghabry:
Currently not used by our code but in the filesystem code.
Buffer was not properly terminated.

And replaced the code with strncpy because that's much easier.


Ghabry:
This only broke some titles because the parser is quite tolerant and accepts "[RPG_RT]�" as a valid section header and Encoding ignores the garbage, too because of atoi which stops when the first non-number is encountered.

Issue #215, opened 2017-10-11
Revise/Improve the CMake build


carstene1ns:
Currently CMake seems not really supported, as it stays in a subfolder.
We should:
- [ ] move `CMakeLists.txt` to the root folder
- [ ] improve the file by removing deprecated/unneeded stuff
- [ ] upgrade Syntax¹

1) We currently use 2.8, however 3.0 has been released 2014, so things could be [tweaked](https://cmake.org/cmake/help/v3.0/release/3.0.0.html#new-features) (e.g. [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html#policy:CMP0048)).

Ghabry:
One 3.x version added FindICU, but couldn't figure out yet which version. This would be the minimal version imo.

carstene1ns:
Changelog says 3.7, which is pretty recent (Nov 2016). I think we should just ship it ourselves.

Ghabry:
Yeah thats, too new. Good option assuming the FindICU script is compatible with 3.0 :D

carstene1ns:
Debian Jessie (oldstable) had 3.0.2, so this should be the absolute minimum for us.
I vote for 3.6, as this is available as backport there and stretch has already 3.7.

Ghabry:
okay than 3.6 is fine for me when even Debian is up-to-date here :D.
Because 3.7 is already so much used I vote against bundling the new FindICU script and to add a conditional instead that checks for < 3.7 and uses pkg-config (as we already do) in that case and for >= 3.7 the new fancy script.

carstene1ns:
Merged the PR, updated the checkmarks.
However, we have now a broken `liblcf-config.cmake`... :see_no_evil:
Build log excerpt:
```
-- Found liblcf: /usr/lib/liblcf.so (found version "0.5.3") 
CMake Error at /usr/lib64/cmake/liblcf/liblcf-config.cmake:88 (target_compile_features):
  target_compile_features called with non-compilable target type
Call Stack (most recent call first):
  CMakeLists.txt:220 (find_package)
```

This breaks here: https://github.com/EasyRPG/liblcf/blob/master/builds/liblcf-config.cmake.in#L88

From https://cmake.org/cmake/help/v3.7/command/target_compile_features.html (emphasis mine):
>The named <target> must have been created by a command such as add_executable() or add_library() and **must not be an IMPORTED target**.

PR #225, opened 2018-03-03
Updated C++11 enabling in cmake

hellow554:
This is the prefered way to enable C++11 support in cmake. This way is also cross plattform and does work with msvc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,13 +27,9 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
 endif()
 
 # C++11 support is needed
-include(CheckCXXCompilerFlag)
-CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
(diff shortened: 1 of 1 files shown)


Ghabry:
thanks for the patch.
There is already an unfinished PR that reworks the CMake script using more modern CMake: https://github.com/EasyRPG/liblcf/pull/224
But "set(CMAKE_CXX_EXTENSIONS OFF)" is missing there.

hellow554:
Uspidaisy... Up to you if merge or closing :)
ALso, do you update the player CMake file as well? Same problem there.

Ghabry:
Yes I'm also reworking the Player CMake file... but I forgot to add the C++11 related lines. ups :D

carstene1ns:
> This is the prefered way to enable C++11 support in cmake.

This works since cmake 3.1. Our configuration targets 2.8 currently. Please _at least_ mention that or bump the version. Not saying we want to stay at this old version, but this is how you break it... :grinning:

carstene1ns:
See #224 and #215. Closing.

PR #227, opened 2018-05-27
Fix two left CMake annoyances

carstene1ns:
Policy 48 is about the new VERSION setting of project(), which will get
confused when including liblcf in Player (not resetting the variables).

We cannot enforce C++11 for liblcf from the configuration file, as we
use an imported target there.


Ghabry:
Oh, sorry for breaking the config file and no idea how I didn't notice this o.O

carstene1ns:
Maybe you did not install after adding this or did you only use internal liblcf from player? Who knows? :face_with_head_bandage:

PR #228, opened 2018-07-06
Exchange State's HP and SP map damage data (val <-> steps)

Albeleon:
Exchange State->hp_change_map_steps and State->hp_change_map_val, and State->sp_change_map_steps and State->sp_change_map_val.

Problem: The data saved in each pair was supposed to be in the other.


Ghabry:
Ping @20kdc

PR #230, opened 2018-07-25
Include stdio unconditionally in reader_lcf.cpp

BSzili:
In reader_lcf.cpp stdio.h only got included when NDEBUG was defined, but it should be the other way around.

### DIFF
--- a/src/reader_lcf.cpp
+++ b/src/reader_lcf.cpp
@@ -8,12 +8,10 @@
  */
 
 #include <cstdarg>
+#include <cstdio>
 #include <istream>
 
 #include "reader_lcf.h"
-#ifndef NDEBUG
-#  include <stdio.h>
-#endif
 
 // Statics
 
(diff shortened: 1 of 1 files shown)


carstene1ns:
I am pretty sure this is intended.

`Read()` vs. `Read0()`: Before was aborting in release builds and printing debug info in debug builds.
With the changed logic now, this will abort in debug build, not printing anything and ignore read errors in release build.

carstene1ns:
Yes, I am fine with the changed include, what I am talking about is the other change.

BSzili:
I see, then I misunderstood you, I see what you meant now. I'll change it to an unconditional include then, that should cover all the cases.

carstene1ns:
jenkins: ok to test!

carstene1ns:
Thanks, hope to see you back when there are AmigaOS/MorphOS/AROS related changes in Player. :grinning:

PR #231, opened 2018-07-27
 Make liblcf compile on MorphOS and AmigaOS4 

BSzili:
The library now builds for MorphOS and AmigaOS4 with autotools.

### DIFF
--- a/src/reader_lcf.cpp
+++ b/src/reader_lcf.cpp
@@ -8,12 +8,10 @@
  */
 
 #include <cstdarg>
+#include <cstdio>
 #include <istream>
 
 #include "reader_lcf.h"
-#ifndef NDEBUG
-#  include <stdio.h>
-#endif
 
 // Statics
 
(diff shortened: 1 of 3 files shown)


carstene1ns:
No ICU support? There will be no encoding detection available...

Ghabry:
Is code for AmigaOS crosscompiled?
For problems in crosscompiling ICU on little endian for big endian you can check the Wii folder in our buildscripts repo (https://github.com/EasyRPG/buildscripts)
We also have patches for removing pthreads from ICU for different versions of it btw.

BSzili:
I was able to build ICU, but as expected it adds a good 26 MB to the executable size. I'd prefer to keep iconv const fix too, as a backup.

carstene1ns:
You likely need to use a smaller data file, by default is very big: [location](https://github.com/EasyRPG/buildscripts/blob/master/shared/packages.sh#L107), [destination](https://github.com/EasyRPG/buildscripts/blob/master/shared/common.sh#L230).
This `ICONV_CONST` thingie was taken care of by an autoconf macro before, but as we migrated to ICU for all supported platforms, we ditched it (just keeping as legacy code).

BSzili:
Ah, that's much better. Now it's only about 1MB larger with all the required ICU libs.

Ghabry:
Long time before the macro was also a boost type trait that fixed the calling by probing the iconv function signature on compile time. Was removed because it was the only remaining boost dependency.
This solution could be also readded because we use now C++11 where this is built in functionality.

Ghabry:
nvm we discussed this already before lgtm

Issue #234, opened 2018-09-25
Reading and writing ldb files should be bit perfect


fmatthew5876:
I'm trying to achieve a bit perfect copy when reading and writing an ldb database. The reason is that I'm using C++ code with liblcf to auto generate stuff in my RPG_RT database. I started going down this path but I found that sometimes liblcf will randomly corrupt the database and then RPG_RT.exe won't load it anymore.

I have a test driver which simply reads the DB and then writes it back out again using liblcf. With this test there are many differences in the resulting file.

I will start submitting PR's to fix these issues as I discover them.

PR #235, opened 2018-09-25
Call Actor::Setup() when writing ldb file

fmatthew5876:
reader_struct doesn't call Actor::Setup() to properly default the initial and final levels. The result is that liblcf writes some bytes where it shouldn't.

This adds a DefaultStruct<T>() wrapper which lets you customize how a default object is created. For most it's just a constructor, but for `Actor` you need to call `Setup()`.

Part of #234

### DIFF
--- a/src/reader_struct.cpp
+++ b/src/reader_struct.cpp
@@ -36,6 +36,22 @@ void Struct<S>::MakeTagMap() {
 		tag_map[fields[i]->name] = fields[i];
 }
 
+template <typename T>
+struct StructDefault {
+	static T make() {
+		return T();
+	}
+};
+
+template <>
+struct StructDefault<RPG::Actor> {
+	static RPG::Actor make() {
+		auto actor = RPG::Actor();
+		actor.Setup();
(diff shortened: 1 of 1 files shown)


Ghabry:
0.5.4 will be released soon (tm), therefore I want to avoid LCF-changes that can break stuff in the Player by accident which is not the case for LDB-writes, therefore I'm fine with merging this at anytime.

Ghabry:
Jenkins: Test this please <3

src/reader_struct.cpp
```diff
@@ -36,6 +36,22 @@ void Struct<S>::MakeTagMap() {
 		tag_map[fields[i]->name] = fields[i];
 }
 
+template <typename T>
+struct StructDefault {
+    static T make() {
```
carstene1ns: liblcf uses tabs, not spaces for indentation.

fmatthew5876:
fixed style nits.

carstene1ns:
jenkins: test this please... </3

Issue #229, opened 2018-07-24
liblcf should not skip empty arrays


fdelapena:
This issue makes liblcf remove chunks which should be saved empty.
This affects starting party and may be related with an old menu commands issue.
Changing this behaviour may require updating stuff in Player.

20kdc:
I don't think fixing this will directly impact Player (saving empty arrays can only make things more correct, short of optional arrays that we don't know about), but giving the correct default values, which I think are:

party: length 1, contents: [1]
menu commands: length 1, contents: [0]

could cause conflict with the 2k3 detection heuristics. The second might, in any case.

20kdc:
...ok, so:

The current behavior of not saving empty arrays that are also the default value is probably correct.

Unfortunately, there are a few arrays which should have default values which don't, and are thus not saved because their default values are empty.

In particular: Chipset's 3 arrays, System party, and System menu\_commands (those last two I noted above: They don't have default value descriptions in the CSV.)

generator.py right now doesn't support default values with arrays in, so here's my shot at it.
https://gist.github.com/20kdc/7b131f083317bbf19c291893092042c5

Tested to compile, but IDK what kind of havoc it causes when running.
Probably at least a little on .flow with pseudo2k3 detection probably being fried.

Also:
`System,party,f,Vector<Int16>,0x16,[1],Array - Short`
`System,menu_commands,f,Vector<Int16>,0x1B,[0],Array - Short - RPG2003`

Ghabry:
good start @20kdc but it will probably also need changes in reader_struct for the "IsDefault" handling. :/

And some lines of code in the Setup() funcs can be removed.

fmatthew5876:
> could cause conflict with the 2k3 detection heuristics. The second might, in any case.

We just need to check whether the DB is 2k or 2k3. If its 2k, we should never write any of the 2k3 arrays.

> Changing this behaviour may require updating stuff in Player.

I don't think so. The player can already load ldb's generated by RPG_RT.exe which have the empty arrays. All this does is make it more compatible.

carstene1ns:
...and then some smart person comes and edits some 2k game in 2k3 - hello `.flow`.

fmatthew5876:
Wow.. Never underestimate your users.. hah

Does anyone know why they did this for .flow? Do some 2k3 features secretly work within 2k's RPG_RT.exe?

Ghabry:
[...]

PR #236, opened 2018-09-25
Add support for writing empty RM2k arrays for Actor

fmatthew5876:
The first commit is part of the other PR, can ignore it here.

This commit creates a `FieldDefaultWriter` mechanism for writing empty arrays. I've implemented support for Actor skills, states, and attributes. 

This is going the approach of manually adding each array. These arrays seems to have all different conventions, so I'm not sure how easy it would be to generate this from a script.
Also, the complete solution would need to check whether it's rm2k or 2k3 mode. If its rm2k mode, the 2k3 arrays must not be written.

Related to #229 and #234

### DIFF
--- a/src/reader_struct.cpp
+++ b/src/reader_struct.cpp
@@ -36,6 +36,22 @@ void Struct<S>::MakeTagMap() {
 		tag_map[fields[i]->name] = fields[i];
 }
 
+template <typename T>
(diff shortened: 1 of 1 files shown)


fmatthew5876:
Actually forget this, lets script it!

PR #237, opened 2018-09-27
Optimized and correct Flags

fmatthew5876:
- Add is2k3 field to flags.csv
- Support writing different 2k vs 2k3 size
    (example: trooppage condition flags)
- Correctly writes empty flags to ldb file
- No more dynamically allocated objects
- Setup union for easy array access of flags in rpg headers
- Binary search static array of tag strings, no std::map.
- Optimzies read/write routines


src/generated/rpg_animationframe.h
```diff
@@ -14,7 +14,9 @@
 
 // Headers
 #include <vector>
+#include <array>
 #include "rpg_animationcelldata.h"
+#include <array>
```
Ghabry: too many <array> includes, iirc you have to handle thsi in "struct_headers" function

src/reader_struct.h
```diff
@@ -595,9 +595,10 @@ template <class S>
 class Flags {
 public:
 	struct Flag {
-		Flag(bool S::*ref, const char* const name) : ref(ref), name(name) {}
+		Flag(bool S::*ref, const char* const name, bool is2k2) : ref(ref), name(name), is2k3(is2k3) {}
```
Ghabry: is2k2

Ghabry:
Jenkins: Test this please <3

Nice changes, also looks harmless concerning pre-0.5.4 merging 👍

fdelapena:
Jenkins: ok to test

Ghabry:
The flags look correct for me. SavePicture is a special case: This is a new feature of RPG Maker 2003 (English release) but there is no way to detect this version inside the database alone.
Though I don't think that marking it as 2k3 will have negative side effects.

fmatthew5876:
Rebased onto style fixed actor changes.

fdelapena:
Thanks!

We've added you to the Player Team, this should whitelist automatic pull request building in Jenkins CI for multiple EasyRPG repositories. At your option, you may keep the team private if you don't want to display it in your GitHub Profile.

fmatthew5876:
I don't know how I got it into my head that the flag names were sorted.

I've removed the binary search in favor of a simple linear search. This is probably still faster than std::map because N is small. map has really bad locality and wastes a lot of memory storing pointers.

fmatthew5876:
Rebased onto master

generator/templates/rpg_header.tmpl
```diff
@@ -48,9 +48,20 @@ namespace RPG {
...
 		struct Flags {
-		{%- for flag in flags[struct_name] %}
-			bool {{ flag.field }} = {{ field|flag_set(loop.index) }};
-		{%- endfor %}
+			union {
+				struct {
+				{%- for flag in flags[struct_name] %}
+					bool {{ flag.field }};
+				{%- endfor %}
+				};
```
carstene1ns: gcc dislikes these anonymous structs, btw.
`warning: ISO C++ prohibits anonymous structs`

https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html

fmatthew5876:
I pushed a new commit to disable gcc warnings about C11 anonymous structs.

Ghabry:
The problem is now that these pragmas make Visual Studio / MSVC extremely spammy with "unknown pragma" warnings.

So the pragma must be guarded with "#ifndef _MSC_VER" ... 😥

fmatthew5876:
Maybe we can just not compile with `-pedantic` with autotools? `-Wall` should be enough.

Ghabry:
hm yeah, I would slowly prefer removing -pedantic from the jenkins job ;)

fmatthew5876:
Well I put the ugly guards in. So let me know what you decide. I can remove them or leave them there.

fmatthew5876:
Rebased onto master

PR #238, opened 2018-09-27
Replace the Boost preprocessor stuff with a jinja2-template

Ghabry:
Because our templates are now not simple string replacements anymore but scriptable.

This makes the code more verbose but it improves the readability imo because the complexity behind the templates is not hidden anymore.


fmatthew5876:
I find this much more readable and it's always good to remove boost dependencies. No reason to use macro tricks when we're already doing python/jinja2 codegen.

Ghabry:
Reran my test scripts, the results stay the same.

Ghabry:
hmpf, why does this need a rebase again :/

fmatthew5876:
Any reason not to merge this? I can rebase my stuff as needed.

PR #239, opened 2018-09-28
Fix warnings with LCF_DEBUG_TRACE

fmatthew5876:
Fixes a warning with `vector<bool>` bit reference and `%d` in printf.

### DIFF
--- a/src/reader_struct.h
+++ b/src/reader_struct.h
@@ -181,7 +181,7 @@ struct Primitive<std::vector<T>> {
 		typename std::vector<T>::iterator it;
 		printf("  ");
 		for (it = ref.begin(); it != ref.end(); ++it) {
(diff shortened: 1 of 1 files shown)


src/reader_struct.h
```diff
@@ -181,7 +181,7 @@ struct Primitive<std::vector<T>> {
 		typename std::vector<T>::iterator it;
 		printf("  ");
 		for (it = ref.begin(); it != ref.end(); ++it) {
-			printf("%d, ", *it);
+			printf("%d, ", (int)*it);
```
carstene1ns: I like C-style casts, however, as liblcf is C++, there should be a fancy, type-aware c++ variant of a cast, maybe.

fmatthew5876:
Replace the C cast with static_cast

PR #240, opened 2018-09-28
Add Is2k3 to Field class and fields.csv

fmatthew5876:
This adds a special is2k3 flag to fields. The flag indicates that this field is not supposed to be present in a 2k database.

This will be used in further PR's to write empty values correctly to ldb files and not write 2k3 fields to a 2k database.

There is already a bug where Terrain::footsteps is written to 2k databases. That is fixed in this changeset.

The flag could also be used in detection heuristics. For example a simple count of how many is2k3 fields exist in a given ldb file.


fmatthew5876:
Added commit to not write fields to a 2k db if they are 2k3.

generator/csv/fields.csv
```diff
@@ -1,1114 +1,1114 @@
...
-Item,occasion_field1,f,Boolean,0x25,False,Flag
-Item,ko_only,f,Boolean,0x26,False,Flag
-Item,max_hp_points,f,Int32,0x29,0,Integer
-Item,max_sp_points,f,Int32,0x2A,0,Integer
-Item,atk_points2,f,Int32,0x2B,0,Integer
-Item,def_points2,f,Int32,0x2C,0,Integer
-Item,spi_points2,f,Int32,0x2D,0,Integer
-Item,agi_points2,f,Int32,0x2E,0,Integer
-Item,using_message,f,Int32,0x33,0,Integer
-Item,skill_id,f,Ref<Skill>,0x35,1,Integer - RPG2003
```
carstene1ns: So this one was wrongly identified?

generator/csv/fields.csv
```diff
@@ -1,1114 +1,1114 @@
...
-SavePicture,current_magnify,f,Double,0x07,-1.0,double
-SavePicture,current_top_trans,f,Double,0x08,0.0,double
-SavePicture,transparency,f,Boolean,0x09,False,bool
-SavePicture,current_red,f,Double,0x0B,-1.0,double
-SavePicture,current_green,f,Double,0x0C,-1.0,double
-SavePicture,current_blue,f,Double,0x0D,-1.0,double
-SavePicture,current_sat,f,Double,0x0E,-1.0,double
-SavePicture,effect_mode,f,Int32,0x0F,0,int
-SavePicture,current_effect,f,Double,0x10,0.0,double
-SavePicture,current_bot_trans,f,Double,0x12,0.0,double
```
carstene1ns: Seems these have been doubles before at least.

generator/csv/fields.csv
fmatthew5876: Yes, items in rm2k can trigger skills.

generator/csv/fields.csv
fmatthew5876: Excel clobbered the doubles, I was using it to make these changes.

There is also a code change in `generate.py` here to always write doubles as a double constant in the C++ code regardless of whether the csv has a decimal point or not.

fmatthew5876:
> Why the change in casing of booleans True → TRUE

Excel messing things up again. I can fix this back to True if you like.

fdelapena:
> Excel messing things up again.

I'd prefer to keep it with existing Excel changes, then any later update with Excel will keep the diff simpler to check. Because the new column changes the whole file it is a good moment to leave these existing text case committed changes as is.

generator/csv/fields.csv
carstene1ns: Yep, saw that change of course, that's why I made the comment... ;-)

carstene1ns:
Yes, have to agree with @fdelapena here. Excel is unfortunately still the defacto standard tool for such things and our generator simply ignores it.
Though generally I prefer to make the tools adapt to the data and not the data to the tools (e.g. editing code and forcing a different tab size or converting tabs to spaces - eww).

carstene1ns:
Since the file was converted to windows line breaks (btw. #237 changed `flags.csv` too :/) and so had to be "fixed" anyway, I wrote a simple linter in ruby which fixed this and the other style nits. I think about creating an issue about how certain parts of these csv's should be handled. Also we will add some `.gitattributes` to tell git to do line break handling here.

PR #241, opened 2018-09-30
Add EnumTags class and generate tags for all enums 

fmatthew5876:
Introduces an enum_tags<E,N> class which provides enum to string functionality. Generator creates these for all RPG enums.


src/enum_tags.h
```diff
@@ -0,0 +1,85 @@
+/*
```
carstene1ns: Please add this file to [Makefile.am](https://github.com/EasyRPG/liblcf/blob/master/Makefile.am#L115). This is especially needed for distribution archives.

fmatthew5876:
Added to Makefile.am

Ghabry:
Sorry for this basic question but how am I supposed to use this? The template looks pretty straight-forward but I'm lost.

As an example lets say I want a tag-iterator RPG::Item::Type. (accessing kTypeTags directly doesn't work, gives linker error)

fmatthew5876:
> Sorry for this basic question but how am I supposed to use this? The template looks pretty straight-forward but I'm lost.
> 
> As an example lets say I want a tag-iterator for RPG::Item::Type.

You can use it like this:

```c++
for (auto* x: RPG::Skill::kTypeTags) {
    std::cout << x << std::endl;
}
```

>  (accessing kTypeTags directly doesn't work, gives linker error)

My fault. I forgot to provide definitions for these. static variables require a definition, although a lot of times you can get away without one in release builds.

PR #242, opened 2018-09-30
Support bit perfect reading and writing of RPG_RT.ldb

fmatthew5876:
This series of changes supports bitperfect reading and writing of RPG_RT.ldb binary databases. It supports 2k and 2k3 mode.

Closes out issues #234 and #229.

This one depends on all my other PR's.

I've tested it with the following:
* An empty 2k3 database
* The default New project database for 2k3
* An empty 2k database
* The default new project database for 2k
* HH3 (2k game ~ 16MB RPG_RT.ldb)

Please if anyone has time. Checkout this branch, download some rm2k/rm2k3 games and test it against their databases. Let me know if any game doesn't come through perfectly.


CMakeLists.txt
```diff
@@ -113,6 +113,7 @@ add_library(lcf
 	src/generated/lsd_savevehiclelocation.cpp
 	src/generated/rpg_chipset.cpp
 	src/generated/rpg_mapinfo.cpp
+	src/generated/rpg_system.cpp
```
carstene1ns: Please add this also to [Makefile.am](https://github.com/EasyRPG/liblcf/blob/master/Makefile.am#L26). This will fix the test failure as well.

fmatthew5876: Added

fmatthew5876:
.flow would likely fail to convert cleanly as it has a 2k3 database but we detect it as 2k. I don't think there is anything we can do in situations like that.

Regarding Makefile.am, is autotools support really still needed now that we have Cmake?

carstene1ns:
As said, we currently use the autotools for the distribution archives and some packagers will also depend on them (i.e. macOS homebrew, some GNU/Linux distributions, platforms without cmake support).

Ghabry:
but what we will remove soon are the Visual Studio project files which are by far the worst to maintain with there XML bloat.

Ghabry:
@fmatthew5876 
Question about LDB read-write: Did you test this with LDB->XML->LDB or LDB read->LDB write?

fmatthew5876:
> 
> 
> @fmatthew5876
> Question about LDB read-write: Did you test this with LDB->XML->LDB or LDB read->LDB write?

Not yet. I'll look into this later today.

Ghabry:
Okay then I maybe misunderstood what I'm supposed to test.
Because you wrote "bit perfect reading/writing" I expect that a file loaded with LDB_Reader::Load followed by LDB_Reader::Save matches but when I diff e.g. your HH LDB-Database I get "binary files differ".

Ghabry:
Okay, what I did: Compiled liblcf on branch "ldb_enums2", then compiled the following program:
```cpp
#include <cstring>
#include <string>

#include "ldb_reader.h"

int main(int argc, char** argv)
{
        std::string infile;
        std::string outfile;

        infile = argv[1];

        outfile = argv[2];
        LDB_Reader::Load(infile, "1252");
        LDB_Reader::Save(outfile, "1252");

        return 0;
}

```

Then I convert via this tool and run "diff RPG_RT.ldb RPG_RT2.ldb"

> binary files dffer

Also tested it with your HH3 ldb

fmatthew5876:
My test driver is this:

https://github.com/fmatthew5876/hh3-rm2k/blob/fixes-2.1/dev/autodb/main.C

without the call to `doWeaponCopies()`

One difference I see is I'm passing `""` as the encoidng, where you're using "1252"

Ghabry:
okay it doesn't matter if I do LDB->XML->LDB or LDB->LDB, both give the same results.

With your hh3 version I get "identity" but for anything else I fire at the converter the result differs.

Ghabry:
For my test data I get a perfect match for:

For 2k: 14 out of 1110
For 2k3: 66 out of 435

fmatthew5876:
I found a bug in how we read `RPG::Parameter` when `actor.final_level` is not default.

With this fix I've tested:
* hh3 v2.03
* hh3 v2.04
* Alter AILA Genesis
* Dungeoneer

Could you run your test suite again and tell me a couple sample games that still fail?

Ghabry:
For 2k: 5 out of 1110
For 2k3: 60 out of 435

One game is e.g. Witch Heart (http://www.vgperson.com/), didn't test the other vgperson games.

For other tests, well, anything on rmarchiv.tk, e.g. these 19 parts of "Das Pferd am Herd": https://rmarchiv.tk/search/title/asc/Das%20Pferd%20am%20Herd

Ghabry:
For a gigantic LDB of 26 MB (lots of battles) take a look at Heros Realm. Another one, this time with thousands of common events (11 MB) is "VH1 / Violated Heroine" (NSFW).

fmatthew5876:
I fixed some fundamental issues that should hopefully fix a lot of games.

fmatthew5876:
More fixes in. One should hit a lot of rm2k3 games.

Ghabry:
> Is there somewhere you can bulk download these games from one place?

Not yet, currently my "test runner" is an ugly fish-shell for-loop combined with some grep'ing. Have no nice way to copy the LDBs.

Will try to write a small ruby-script for the testing which gives me more possibilities. Then I can send you all the mismatched LDBs.
Will also see if I can get the script working on Jenkins as a test runner :). Going to talk with the other devs about this.

fmatthew5876:
That sounds great.

In the meantime could you run your script again and make a new pastebin against the latest changes? We should be getting very close now.

Ghabry:
You are getting really close now:

2k: 1058 out of 1110 (95%)
2k3: 295 out of 435 (68%)

https://pastebin.com/GQf4rZGZ

fmatthew5876:
Another handful of fixes are in. @ghabry could you run your script again please? :)

I'm seeing a lot of games now with corruption or other weird cases we won't be able to handle. I think we're reaching the end here.

One common thing I've seen in some of the remaining 2k games are animations that have 0x8 screen_shake effect chunks. I'm guessing people are copy and pasting animations from 2k3 games and that block gets carried over. It's a 2k3 only field so liblcf won't write it back out. Maybe we should??

src/generated/rpg_terrain.h
```diff
@@ -89,9 +89,9 @@ namespace RPG {
 		int32_t special_lateral_party = 10;
 		int32_t special_lateral_enemies = 5;
 		int32_t grid_location = 0;
-		int32_t grid_a = 0;
-		int32_t grid_b = 0;
-		int32_t grid_c = 0;
+		int32_t grid_a = 120;
+		int32_t grid_b = 392;
+		int32_t grid_c = 16000;
```
Ghabry: Intersting, this could fix battle issues we have in some RPG2k3 games! (all party members overlapping)

fmatthew5876: Those were a huge pain to figure out. Basically I had to keep guessing and edit the ldb binary manually. Then read/write it with rm2k3 to see if the chunk would dissapear.

I had the same thought that maybe this would fix some bugs in Player. Any game with defaulted terrain would be all 0's and have overlapped actors.

Would you like me to cherry-pick this one out into a separate bug-fix PR so you can merge it for 5.4?

fmatthew5876:
> Well when people use the 0x8 chunk by accident for some reason imo it should be copied over if it improves the compatibility :). But only for chunks which are very commonly present.

We can potentially get away with this for any chunk which is not present in the ldb if its defaulted. But for things that are always written we must filter on is2k3.

src/generated/rpg_terrain.h
Ghabry: Yeah, a cherry pick just for this would be great. thanks.

fmatthew5876:
I've gone through the offending rm2k games.

The remaining ones are documented here in #246

One had a new 0x62 chunk in system. I saw we already have "Unknown_61" so I added "Unknown_62". Does anyone have more background on this?

I saw a few german games with really messed up animation data. In Devil and Angel, an animation chunk is not properly terminated with a 0, so liblcf keeps reading and screws things up badly. This animation also looks corrupt in rm2k editor, but rm2k does load game succesfully. A few others looked like they have the same pattern but I haven't dug deeper. 

Has anyone seen this animation issue before? Was there ever any third party tool for animations that might have caused such a bug?

Will now look at the 2k3's.

fdelapena:
A new chunk after long time without finding, yay!

When an unknown chunk is detected, liblcf should output a console warning. Before this change, it should appear after parsing.

Because both 0x61 and 0x62 are directly after battle test chunks, they might be actually part of battle test parameters. In case they are related, battle test data is set the first time when launching battle test from database: it launches directly from database RPG_RT and passes a special temporally saved RPG_RT.ldb (this makes sense because it is still not saved until the user presses OK in database editor) with current settings. When this gets passed to RPG_RT.exe, it launches it with a special command line interface arguments: BattleTest x x n where n is the ID of the monster group. Once the battle test ends, it returns to database editor window and restores the unsaved RPG_RT.ldb, however once the database gets saved by pressing OK, these chunks get actually stored.

fmatthew5876:
I added an optional argument to `LDB_Reader::Save` to allow you to choose to preserve the non-standard ldb header or not.
```c++
LDB_Reader::Save(outfile, encoding, LDB_Reader::SaveOpt::ePreserveHeader)
```

@Ghabry Could you run your test again with `ePreserveHeader` to confirm my results?

Thanks

fdelapena:
> I added an optional argument to `LDB_Reader::Save` to allow you to choose to preserve the non-standard ldb header or not. The default behavior is to write the standard header.

Cool feature, thanks!

The protection utility also modifies lmt and lmu files, but not sure if suitable for the scope of this PR

fmatthew5876:
> The protection utility also modifies lmt and lmu files, but not sure if suitable for the scope of this PR

LMU, LMT, and LSD all need the same treatment as LDB is getting here. That should be separate PR.

fmatthew5876:
Regarding Jenkins testing and the Terms problem. For games that fail with these, we could convert them once to standardize the terms strings, and then use the copies for testing against future PR's and changesets.

Ghabry:
ok, will have time in a few hours for testing :). Will cherry-pick #250 in my test to check if this fix has unexpected side effects.

Ghabry:
2k: 1084 out of 1092 (99%) 🎉
2k3: 384 out of 405 (95%)
2k3e: 45 out of 45 (100%) 🎉

```
BloodmoonEnt. & HangmanProd
Der Sinn des Lebens (Demo 2) <-- Known as corrupted
Devil or Angel <-- Known as corrupted
Fate Hunter Orleon (Demo)
Kroetenstuhl_V040 <-- Known as corrupted
Krötenstuhl Saga (Demo 5a) <-- Known as corrupted
Rick Jones Hunters (Demo) <-- Known as corrupted
Sweet Dreams (Demo)

2k3

Aliens - This Time it's War (Demo)
Antagonism
Bloody Bastard's Killer Instinct v1.02
Budubi's Scary School
Budubis Scary School (Final Mix)
Der Bote des Königs (Demo)
Die Abenteuer der MS White
Dragonball V
Ebenb�rtig vv1.0 - Noa, White Cat Kyuba - rm2k3 - 30.09.2008�rtig vv1.0 - Noa, White Cat Kyuba - rm2k3 - 30.09.2008�rtig vv1.0 - Noa, White Cat Kyuba - $
Echoes Of The Past
Im Tal des Schreckens 3 (Demo)
In den Wolken
Koma Klaus - Arcade Edition
Mario & Luigi - Lugmillas Return (Demo 3)
Mario's Other Brother 2
Monster World RPG <-- Known as corrupted
Seven Blades - Darkness (v0.44) <-- Known as corrupted
Space Drop - the hero myth (Demo)
Space Drop - the hero myth
Wir Beide
~Haematom
```

fdelapena:
> Does there exist any large cache of saved games somewhere for LSD testing? Getting LSD right could expose bugs in the save/load system.

If I recall correctly, Ghabry preserves a lot of .lsd files from email bug reports send from Android users. The Android port contains a send email bug report feature. However, savegames are not converted from any RPG_RT version but created from scratch, so maybe these are not useful for bit comparison for non-moveable stuff.

For moveable things, .lsd files contain an excel/delphi style timestamp, apart of other living stuff, so for this particular format a lot of bits may vary even from RPG_RT itself. There are some games being distributed with .lsd files, the latest I recall is .flow English translation 0.192.

A way to try getting crowdsourced generated savegames from RPG_RT.exe might be done by sending a message to RMN forums and maybe a twitter announcement from the easyrpg account.

----

A fact about savegames, there is an interesting fact on event execution behavior: there are save count (0x5B) references here, but this one should not increase for every time saved, instead should match with map save count (not sure about ldb 0x5b interaction). [This is taken into account in Player](https://github.com/EasyRPG/Player/blob/master/src/game_map.cpp#L261-L263):

> 0x5B | マップの保存回数 | 整数 | セーブデータロード時にここが変更されていると、「イベント開始条件」が「定期的に並列処理する」のイベントを最初から実行する

Machine translation says:

> 0x5B | Number of save times of map | integer | If this is changed when loading saved data, the event start condition executes the event "start processing in parallel at regular intervals" from the beginning

Which seems to restart event execution like when the player enters the map instead of retrieving event execution position (which has some issues in Player currently). This is taken from some German forums but the original seems from (now broken) http://rpg2kdev.sue445.net wiki, where most lcf information came (including around the 25% of the .lsd format).

I guess this is on purpose because when a map gets modified, loading a saved game from that map may run into unexpected behaviour.

fmatthew5876:
> A way to try getting crowdsourced generated savegames from RPG_RT.exe might be done by sending a message to RMN forums and maybe a twitter announcement from the easyrpg account.

That's a great idea. It possible we could fix a lot of bugs and improve compatibility if we had lots of save files to test.

> 
> There is an interesting fact on event execution behavior: there are save count (0x5B) references here, but this one should not increase for every time saved, instead should match with map save count (not sure about ldb 0x5b interaction). [This is taken into account in Player](https://github.com/EasyRPG/Player/blob/master/src/game_map.cpp#L261-L263) but it seems only when saving, not when loading, or at least I didn't found it:
> 
> > 0x5B | マップの保存回数 | 整数 | セーブデータロード時にここが変更されていると、「イベント開始条件」が「定期的に並列処理する」のイベントを最初から実行する
> 
> Machine translation says:
> 
> > 0x5B | Number of save times of map | integer | If this is changed when loading saved data, the event start condition executes the event "start processing in parallel at regular intervals" from the beginning
> 
> Which seems to restart event execution like when the player enters the map instead of retrieving event execution position (which has [some](https://github.com/EasyRPG/Player/issues/757) [issues](https://github.com/EasyRPG/Player/issues/1075) in Player currently). This is taken from some German forums but the original seems from (now broken) http://rpg2kdev.sue445.net wiki, where most lcf information came (including around the 25% of the .lsd format).
> 


Wow, so that's how this feature is implemented. 

In hh3 I ran into a lot of bugs with my custom menu system related to this. I have to turn off the switch that activates the CMS while the player is accessing the `OpenMenuScreen` and turn it back on immediately after. If the map changed, `RPG_RT.exe` would load the game with the CMS switch turned off and the player couldn't access the menu. Eventually I had to ban the use of `OpenSaveMenu` and wrap it in a common event to work around even more bugs triggered by this.

> I guess this is on purpose because when a map gets modified, loading a saved game from that map may run into unexpected behaviour.

Yeah that's exactly why. We would need to emulate something like this for editor. Also for people taking old saves and playing them with newer versions of games. We could load the save improperly and be incompatible with rm2k/3.

What about 2k3 steam games that use 0x5A for save game counts? Does it still trigger the same behavior off of 0x5A? Does lsd use 0x5A instead of 0x5B too?

fdelapena:
> What about 2k3 steam games that use 0x5A for save game counts? Does it still trigger the same behavior off of 0x5A? Does lsd use 0x5A instead of 0x5B too?

It seems this was just to prevent the behavior of RPG_RT events. I've seen time ago somewhere some discussion about the decision, but there is also some mention in the 2k3 official English release changelog which broke something in the first official English release (1.10). This is part of the 1.11 changelog:

> Regression from v1.10: When a savegame was loaded after the current map was modified, the events didn’t reset. If event modifications or deletions were part of this change, the internal state could become corrupted and the game would crash when trying to load the savegame. Also, newly created events didn’t show up until the map was left and reentered.

I don't remember if this was changed back again to 0x5B or got some additional patches by Cherry to keep it as 0x5A. Note there is another related chunk in .lsd (map_save_count, 0x83). Ghabry knows better this story.

----

A hidden gem (and not widely announced yet) in Player is a log based [input replay](https://github.com/EasyRPG/Player/pull/1096). This might be useful when regenerating a savegame from scratch in Player using the same game keys sequence, which should be frame precise. There is no problem with games using random features, because Player also supports since a while a deterministic seed command line parameter (that is not possible in RPG_RT, though).

With this feature, generating clean savegames to compare with the previous one may help to check all interpreter execution differences after changing some defaults in lsd save/load generation. Apart of the possibilities in automated testing already mentioned, this might be checked with higher precision thanks to the verbose logs dumped by the amazing [event tracer](https://github.com/EasyRPG/Player/issues/646) feature (which has a counterpart in DynRPG for RPG_RT) to compare with frame precision the step by step event interpreter behavior on savegame load.

Ghabry:
codewise it looks fine to me but I will do a retest with LDB and also add LMT and LMU files. (maybe not all LMU files though because there are so many...)

Ghabry:
Going to bed now, therefore only dumping you the raw output of my improved test script: (omitted LSD as this is part of another PR) and havn't filtered the bogus games you already reported...

For Maps I only used 0001 because I noticed that when the 1st map fails usually all other maps fail, too and using all maps far too long.

http://ix.io/1qkp

fmatthew5876:
I don't see any new regressions for LDB from your list. All the ones there have some corruption or have terms issues.

Sweet Dreams (Demo) all the files (LDB,LMT,LMU) have 9 or 10 extra zero bytes at the end. I don't know what these are supposed to be for but liblcf doesn't write them back out.

This is the only one failing LMT testing, so we can sign off on LMT being done.

For LMU, are you sure your test is correct? I tested some of the failing games in your list and they came out clean, all LMU files.

Do you load the database and the treemap into memory before loading and saving maps? I do that in my test as they can rely on those data structures to parse correctly.

I'm using a hacked up version of the below test driver to do maps. The call to `loadRPG()` in there loads the LDB and LMT.

* https://github.com/fmatthew5876/hh3-rm2k/blob/fixes-2.1/dev/copyrpg/main.C

Ghabry:
My C++ test program is a bit less sophisticated but I'm indeed not loading the LDB or LMT which means the version field is not initialized... my fault. Won't have access to my PC before the evening, then I rerun it. I wonder if it will fix the 2k games, too (and then I wonder why).

I confirm that there are no new regressions which means I will remove all the non-fixable games from my test data.

Ghabry:
hm I messed something up, now I get for LMU
```
Summary:
2k:   1072/1092 (98.2%)
2k3:  402/416 (96.6%)
2k3e: 44/45 (97.8%)
```

but for LDB 
```
2k:   14/1092 (1.3%)
2k3:  404/416 (97.1%)
2k3e: 45/46 (97.8%)
```

investigating...

Issue #249, opened 2018-10-03
Reader: Handle corrupted chunks better


fmatthew5876:
Some games in #246 have corrupted chunks that liblcf can't handle. One example animation chunk from Devil and Angel is not 0 terminated, causing liblcf to continue reading past the end of the chunk and getting confused.

rm2k/3 is able to run these games, so it's a compatibility issue.

Make liblcf able to load these ldbs in the best way possible. More compatible with rm2k/3 the better. The games can be opened in the rm2k/3 editor to compare how liblcf interprets them vs how rm2k/3 does it.

Approach 1:

When returning from `ReadLcf()` calls, check how many bytes were read vs how many should have been read via `LcfSize()`. If the numbers don't match, seek back to the right position before going to the next chunk.

Approach 2:

Add a check before every read operation to ensure we never pass `LcfSize()`. This approach might be slower.

Ghabry:
@fmatthew5876 As #250 is merged would you consider this resolved?

fmatthew5876:
I guess so. We could further try to parse corrupted chunks exactly like RPG_RT does but I see little benefit to that.

PR #250, opened 2018-10-03
Detect corrupted chunks and reset the parser

fmatthew5876:
Towards addressing #249 

This one successfully parses the Devil and Angel ldb. When I copy it, the chunk for the corrupted animation is different, but everything else in the database is a perfect match.

This and other games like like it should load and be playable now in Player.

This doesn't attempt to emulate how rm2k/3 interprets the corrupted chunk.

### DIFF
--- a/src/reader_struct.cpp
+++ b/src/reader_struct.cpp
@@ -73,10 +73,18 @@ void Struct<S>::ReadLcf(S& obj, LcfReader& stream) {
 #ifdef LCF_DEBUG_TRACE
 			printf("0x%02x (size: %d, pos: 0x%x): %s\n", chunk_info.ID, chunk_info.length, stream.Tell(), it->second->name);
 #endif
+			const auto off = stream.Tell();
 			it->second->ReadLcf(obj, stream, chunk_info.length);
+			const auto bytes_read = stream.Tell() - off;
+			if (bytes_read != chunk_info.length) {
+				fprintf(stderr, "Warning: Corrupted Chunk 0x%02x (size: %d, pos: 0x%x): %s : Read %d bytes! Reseting...\n",
+						chunk_info.ID, chunk_info.length, off, it->second->name, bytes_read);
+				stream.Seek(off + chunk_info.length);
+			}
 		}
-		else
+		else {
 			stream.Skip(chunk_info);
+		}
 	}
 }
 


fmatthew5876:
Matches my results. Thanks!

CherryDT:
The RPG Maker does the same.

I found it interesting that there were two games (Devil or Angel and Monster Hunter RPG) with the exact same corruption, but in total different locations in the file. I looked at it and saw that they are both based on the same template, which I believe to be an obscure Italian-translated RM2k default database (I guess Monster World was converted from 2k to 2k3).

My guess is that in the world of corrupted downloads 15 years ago, this version was corrupted once and then it spread in some Italian-translated RM2k download...

The corruption is a 00 which should have been a 03 - that fixes it.
image

I also checked Rick Jones Hunters and there is a single bitflip at a different place, but also inside animation timings (a 09 which should have been a 01):
image

I do wonder though why this seems to happen almost exclusively in animation timing data. I'm worried that there is some kind of bug in RM (like a buffer overflow or something) which can cause these issues infrequently...

Now that's nothing liblcf can do of course, I just found it interesting nonetheless.

Ghabry:
I find it interesting that RPG_RT is able to read such a corruption but when you have one extra trailing 0-byte at the end of a savegame you get "Stream Read Error" m(

fmatthew5876:
> I'm worried that there is some kind of bug in RM (like a buffer overflow or something) which can cause these issues infrequently...

It's possible but I would say unlikely. Out of thousands of games we tested only these few have any corruption at all. I myself used rm2k extensively for 5 years when I was younger and never had any kind of corruption.

I would guess somehow something got corrupted that someone downloaded or shared. Then people copied stuff out of it or used it as a base and so it spread. But really, who knows.

ErasSoft:
> The RPG Maker does the same.
> 
> I found it interesting that there were two games (Devil or Angel and Monster Hunter RPG) with the exact same corruption, but in total different locations in the file. I looked at it and saw that they are both based on the same template, which I believe to be an obscure Italian-translated RM2k default database (I guess Monster World was converted from 2k to 2k3).
> 
> My guess is that in the world of corrupted downloads 15 years ago, this version was corrupted once and then it spread in some Italian-translated RM2k download...
> 
> The corruption is a 00 which should have been a 03 - that fixes it. image
> 
> I also checked Rick Jones Hunters and there is a single bitflip at a different place, but also inside animation timings (a 09 which should have been a 01): image
> 
> I do wonder though why this seems to happen almost exclusively in animation timing data. I'm worried that there is some kind of bug in RM (like a buffer overflow or something) which can cause these issues infrequently...
> 
> Now that's nothing liblcf can do of course, I just found it interesting nonetheless.

Yes, Devil or Angel based on a Italian rpg maker. :) Good work :)

PR #252, opened 2018-10-04
Identify System 0x61 and 0x62

### LINKED ISSUE #248: Figure out System Chunks 0x61 and 0x62
System has chunks 0x61 and 0x62 which are likely part of battle test data. Figure out what these chunks actually do.

Also mentioned here #242:

> When an unknown chunk is detected, liblcf should output a console warning. Before this change, it should appear after parsing.
> 
> Because both 0x61 and 0x62 are directly after battle test chunks, they might be actually part of battle test parameters. In case they are related, battle test data is set the first time when launching battle test from database: it launches directly from database RPG_RT and passes a special temporally saved RPG_RT.ldb (this makes sense because it is still not saved until the user presses OK in database editor) with current settings. When this gets passed to RPG_RT.exe, it launches it with a special command line interface arguments: BattleTest x x n where n is the ID of the monster group. Once the battle test ends, it returns to database editor window and restores the unsaved RPG_RT.ldb, however once the database gets saved by pressing OK, these chunks get actually stored.

fmatthew5876:
System 0x61 can be used by Player for compatible rm2k3 equipment usage.

Resolves #248 

Depends on #240


fmatthew5876:
I studied the behavior of 0x62 a bit.

When you double click on a terrain in Troops, 0x62 changes to that terrain id. 0x54 changes to the battle background. If you select a battle background in Troop page, it changes 0x54 to that and 0x62 goes to 0 and doesn't get written out if you save.

I can't tell what other effect it has on the UI or battle test if any. You can select terrain from the battle test menu already, so I don't know what this feature is supposed to offer.

PR #262, opened 2018-10-12
Fix how SaveGame Music and sfx are loaded

fmatthew5876:
If the chunk name is an empty string, it means default the entire chunk to
ldb system, not just the name. That includes name, tempo, volume, balance, etc..

Fixes an issue in Player https://github.com/EasyRPG/Player/issues/985 where battle start SFX tempo was reset to 100 when loading saved games.


Ghabry:
hm this looked simple enough for 0.5.4, well, too late now :)

PR #264, opened 2018-10-16
Identify remaining unknown chunks

fmatthew5876:
Adds the info from #261 to fields

~~Depends on #242 because rebase hell~~

PR #266, opened 2018-10-17
 Increment save counts when writing LDB, LMU, and LSD 

fmatthew5876:
Depends on: #264

This patch changes *_Reader::Save() to automatically increment `save_count` before writing the file.

There is a new `SaveOpt::eNoIncSaveCount` you can pass to suppress this behavior.

I also added `SaveOpt::eNoUdate` value which is intended to combine all of the `SaveOpt`'s required for copying lcf files without modifications. We should use this `SaveOpt` for our testing liblcf.

This will fix bugs in Player and Editor related to save counts.

### DIFF
--- a/src/ldb_reader.cpp
+++ b/src/ldb_reader.cpp
@@ -17,6 +17,10 @@
 #include "reader_util.h"
 #include "reader_struct.h"
 
+void LDB_Reader::PrepareSave(RPG::Database& db) {
+	++db.system.save_count;
+}
+
 bool LDB_Reader::Load(const std::string& filename, const std::string& encoding) {
 	std::ifstream stream(filename.c_str(), std::ios::binary);
 	if (!stream.is_open()) {
--- a/src/ldb_reader.h
+++ b/src/ldb_reader.h
@@ -42,6 +42,11 @@
  */
 namespace LDB_Reader {
 	/**
+	 * Increment the database save_count.
+	 */
+	void PrepareSave(RPG::Database& db);
+
+	/**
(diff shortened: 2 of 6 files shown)


Ghabry:
The question is if incrementing the save count should be the default option or not.
One will need changes to Player and editor, the other to all other tools working with LCF files.

btw for save games there is also a timestamp. The Player currently handles this by setting the timestamp but this could be also done via an option.

fmatthew5876:
> The question is if incrementing the save count should be the default option or not.
> One will need changes to Player and editor, the other to all other tools working with LCF files.

Good question. It's also wierd that now the `Save_*` functions now take the object by non-const reference.

The only thing I feel strongly about is that the increment logic should be in `liblcf` so it can be written once and shared. Maybe a separate function that needs to be called explicitly by Player and Editor before saving?

fdelapena:
Jenkins: test this please

Ghabry:
Doesn't really matter for me if

- individual function in liblcf or
- invert the logic of the enums because then we only have to update Player and Editor and keep anything else untouched.

SaveOpt::Update, SaveOpt:IncSaveCount and SaveOpt::UpdateTimestamp (this is for the timestamp in the LSD, currently done in Player but the timestamp function is in liblcf).

Originally liblcf always updated the timestamp when saving but this was obviously bad when comparing files :)

fmatthew5876:
New implementation. IMO this is much cleaner.

fdelapena:
Jenkins: test this please

PR #271, opened 2018-10-30
Add Maniac Patch Event commands

Ghabry:
SetMousePositon surprisingly doesn't add a new chunk to the savegame, which means the offset is not stored.

Not fully sure what "EndLoading" is supposed to do, doesn't match with the documentation. waiting for a response on twitter.

### DIFF
--- a/generator/csv/enums_easyrpg.csv
+++ b/generator/csv/enums_easyrpg.csv
@@ -1,0 +1,17 @@
+#Structure,Entry,Value,Index
+EventCommand,Code,Maniac_GetSaveInfo,3001
+EventCommand,Code,Maniac_Save,3002
+EventCommand,Code,Maniac_Load,3003
+EventCommand,Code,Maniac_EndLoadProcess,3004
+EventCommand,Code,Maniac_GetMousePosition,3005
+EventCommand,Code,Maniac_SetMousePosition,3006
+EventCommand,Code,Maniac_ShowStringPicture,3007
+EventCommand,Code,Maniac_GetPictureInfo,3008
+EventCommand,Code,Maniac_ControlBattle,3009
+EventCommand,Code,Maniac_ControlAtbGauge,3010
+EventCommand,Code,Maniac_ChangeBattleCommandEx,3011
+EventCommand,Code,Maniac_GetBattleInfo,3012
+EventCommand,Code,Maniac_ControlVarArray,3013
+EventCommand,Code,Maniac_KeyInputProcEx,3014
+EventCommand,Code,Maniac_RewriteMap,3015
+EventCommand,Code,Maniac_ControlGlobalSave,3016
--- a/generator/generate.py
+++ b/generator/generate.py
@@ -417,7 +417,7 @@ def main(argv):
 
     structs = get_structs('structs.csv','structs_easyrpg.csv')
     sfields = get_fields('fields.csv','fields_easyrpg.csv')
-    enums = get_enums('enums.csv')
+    enums = get_enums('enums.csv','enums_easyrpg.csv')
     flags = get_flags('flags.csv')
     setup = get_setup('setup.csv')
     constants = get_constants()
--- a/src/generated/rpg_eventcommand.h
+++ b/src/generated/rpg_eventcommand.h
@@ -153,7 +153,23 @@ namespace RPG {
 				EndLoop = 22210,
 				Comment_2 = 22410,
 				ElseBranch_B = 23310,
-				EndBranch_B = 23311
+				EndBranch_B = 23311,
+				Maniac_GetSaveInfo = 3001,
+				Maniac_Save = 3002,
+				Maniac_Load = 3003,
+				Maniac_EndLoadProcess = 3004,
+				Maniac_GetMousePosition = 3005,
+				Maniac_SetMousePosition = 3006,
+				Maniac_ShowStringPicture = 3007,
+				Maniac_GetPictureInfo = 3008,
+				Maniac_ControlBattle = 3009,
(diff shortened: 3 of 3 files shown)


carstene1ns:
I think this should involve a generator change that applies some "Maniac Patch" comment here.

Ghabry:
jadehpy wrote me that End Loading reduces the blackscreen after load from 6 frames to 1 frame but the blackscreen here is around 0.5s long, so I can't really notice this 5 frame difference o.O.

fmatthew5876:
I agree with carstene1ns, these need to clearly be marked as non-standard in some way.

Also #272

Ghabry:
I agree with your objections.

Note that this patch fulfills the "doesn't conflict with normal games" policy, otherwise I wouldn't even consider adding it.

fmatthew5876:
Maybe even a separate `fields_maniac.csv` or something like that?

Ghabry:
I find these commands so useful that I would even consider them for the EasyRPG Editor.
Which spawns the discussion where to put (potentially) new Event commands invented by us.

In this case ``enums_maniac.csv`` (no new fields) and ``*_easyrpg.csv`` feels more future proof.

fmatthew5876:
`*_easyrpg.csv` makes sense to me for custom stuff.

We need to decide on some uber-format to support all patches, but there is no reason we can't use exactly the maniac convention to start with. If some other patch implements the same thing we'll just have to translate it at the liblcf level.

fmatthew5876:
For `EventCommand::Save`, what is the value of `SaveSystem::screen` in some of those maniac saved games?

Normally in `RPG_RT`, all saving goes through the save menu, and therefore `screen == 5` (filemenu).

If this command lets you save anywhere, I'd bet `screen == 0` (map). Can you save in battles?

Ghabry:
new solution. opinion?

fmatthew5876:
My vote is for `enums_easyrpg.csv` instead of having separate files for different patches.

In other words, we have base `RPG_RT` compatibility layer, and then our easyrpg customization layer which also incorporates support for patches (with data transforms if needed).

Ghabry:
I'm okay with this solution 👍

fmatthew5876:
Apologize for being so nitpicky, but if they are `easyrpg_enums.csv` then I think the enum tags themself should also be easyrpg and not `Maniac_`.

How about an `EZ_` prefix? Like `EZ_Save`? Or `EasyRPG_Save` if you want to be longwinded.

Ghabry:
Well the problem is that these event commands are specific to Maniac Patch and we decided to only have one file.

I would be fine with "EZ_", doesn't really matter for me.
Waiting for @fdelapena and @carstene1ns favourite, than I will choose the winner. xD

fmatthew5876:
To give more context of where I am coming from, I'm thinking to make a PR add these custom equipment settings. This is why I made `equipment_setting` an enum and not a bool.

```
EZ_EquipmentSetting_either
EZ_EquipmentSetting_both
```
(where EZ can be whatever convention we decide)

That would be a pure EasyRPG extension.

fdelapena:
I agree once EasyRPG Editor adopts this feature as own, keeping the patch name prefix may not be worth, adding a comment or documentation somewhere about the origin and design decision of those numeric IDs is still useful, though.

Ghabry:
soooooo, what is the result? Do you want Maniac, EZ or Easy-prefix now?

carstene1ns:
Since all of these additions are good ideas to support later, should be okay to use our prefix, but may be a good idea to clarify the origin with the current maniac prefix.

fmatthew5876:
Why don't we have 2 conceptual name spaces? 

`EasyRPG` and `Maniac`. 

Each custom command that's compatible with a patch should have alias in that patch's namespace. It also needs a home in the EasyRPG namespace in order to be usable by our tools.

This will be the foundation of logic needed later to convert incompatible map patch customization EasyRPG.

fmatthew5876:
Regarding this save command. How does it work? Does it save immediately or does it wait till the end of the frame and save at the same time opensavemenu would be called?

If it's the former then I don't support adding this command. It's useful for debugging and research for sure, but there are many local variables mid frame which don't have LSD chunks.


One that comes to mind is the loop counter in the interpreter that goes up to 10000.

The result will be heisenbugs with broken saves done at the wrong time. I would never use such a feature in a game.

Ghabry:
That's a good point, unfortunately there is no event tracer for 2k3 Steam due to lack of DynRPG to figure this out easily.
I would expect an interpreter yield, otherwise this would be really a bad design choice, will try to figure this out.

Ghabry:
While testing this I directly found a bug with the normal scene calling:

Parallel Process 1:
```
Loop
 V1 += 1
 if V1 > 1000
  Save
 endif
Repeat
```

PP2:
```
Message: V[1]
```

A blank non-transparent message box appears right before it fades to the save scene and after closing the scene it draws the text. The next boxes are all correct. Must be a problem with initialisation and fade, is fixed by a Wait 0.0.

@fmatthew5876 You were right.
I simply use the "execution limit" of Parallel Processes to test this: I save and look at the message box after loading:
The output is 3000 (Maniac) vs 5000 (Scene) which means it doesn't use the same way because the output is different.

I still like this command but we can't support it as fully intended (but was likely just not properly thought about a said)

fmatthew5876:
Here's another,

Parallel events are allowed to continue running while a message box is up. So you could save your game while a message box is active.

I'd like to see what happens when you try to load that save..

Ghabry:
Finally rebased it and added all commands up to version 190920

PR #275, opened 2018-11-01
Add cmake config to autotools-based installs

carstene1ns:
Fixes an issue where an old cmake version (<3.10) will not compile
Player, because the needed EXPAT::EXPAT target is not available.


Ghabry:
I looked at other libraries that bundle Find modules and the solution is using the "include" directive. No idea if this has any unwanted side-effects but for me it works (also added a <3.11 version check as starting with 3.11 EXPAT was added.)

carstene1ns:
Cannot approve my own pull request `*g*` :+1:

Ghabry:
haha yes, couIdnt add you as a reviewer neither.

PR #268, opened 2018-10-18
Bitperfect copying of LSD

fmatthew5876:
Fixes #246

This series of changes is towards doing perfect copies of LSD files.

So far I am able to copy a few local save games I have. 

If anyone has lots of saved games, please test copying with this branch of send the saves to me.

PR #277, opened 2018-11-02
SaveGame Fixes

fmatthew5876:
Depends on #268 

This PR is attempting to fix how LSD files are written from Player. A lot of this includes proper sizing of arrays and clearing out chunks which have the same value as in the LDB / LMU.

The way I'm testing this is:
1. Create a very simple test game
2. Start New Game and Save in `RPG_RT.exe`
3. Start New Game and Save in `Player`
4. Binary compare the resulting lsd files and track down differences.
5. Start adding things to test game, and repeat the process.

Most of this work will end up in liblcf, but some changes are needed in Player too. The corresponding Player PR is https://github.com/EasyRPG/Player/pull/1470

More work is needed here, but already I've come across some serious bugs in how we use save files and some unknown fields.

@CherryDT

What are these fields?
* SaveSystem::screen 0x1
* SavePartyLocation::stop_count 0x34

One very interesting problem is `Save::pictures`. For a new game, this array needs to be resized to exactly the number of a picture that this particular version of `RPG_RT.exe` supports.

Number of pictures changed quite a lot. I don't know all of the different versions and what their picture counts were.

Does anyone know:
* What versions of rm2k and rm2k3 had picture count changes?
* Can we accurately detect all of those versions?


Ghabry:
Not Cherry but I can guess most :)

screen is the ID of the screen where you are currently. Doesn't matter at all for loading afaik, therefore would simply write "Map" there, see: http://rpg-maker.cherrytree.at/dynrpg/namespace_r_p_g.html#a2385fd16f4ad88a4166fc9d04255a88c

stop_count is the "stop_count" of Main_Data::game_player

anim_count is the animation frame of Main_Data::game_player

panorama_data does not export any savable chunks and is unused.

frame_count: Probably frames since startup? Don't think it's important to get this right.

Ghabry:
Is the picture count really relevant? RPG_RT doesn't really care what you put there, also there are patches which adjust the amount of pictures, is not really possible to guess the versions.
Imo it is good enough when RPG_RT can load the save correctly instead of adding complexity for corner cases just because of "RPG_RT does it this way" (for all other LCF formats I agree with this, but LSD?)

src/rpg_fixup.cpp
```diff
@@ -129,8 +168,107 @@ void RPG::SaveSystem::Fixup() {
 	}
 }
 
+void RPG::SaveSystem::UnFixup() {
+	const RPG::System& system = Data::system;
+
+	//TODO: Should be a C++14 polymorphic lambda
```
Ghabry: what is a polymorphic lambda?

src/rpg_fixup.cpp
```diff
@@ -168,6 +168,18 @@ void RPG::SaveSystem::Fixup() {
 	}
 }
 
+template <typename T>
+static void trimDefauled(std::vector<T>& v, T def) {
```
Ghabry: Defauled

generator/csv/fields.csv
```diff
@@ -876,10 +876,10 @@ SavePartyLocation,unboarding,f,Boolean,0x68,False,0,0,
...
 SavePartyLocation,menu_calling,f,Boolean,0x6C,False,0,0,Flag which briefly is true if the player presses ESC. At the right place in handling each frame's activities for the player; the code checks whether this flag is set and calls the menu; however there are several conditions which would cancel this flag and instead process another higher-priority action; such as when an encounter takes place during the same frame.
 SavePartyLocation,pan_state,f,Int32,0x6f,1,0,0,
-SavePartyLocation,pan_current_x,f,Int32,0x70,2304,0,0,int
-SavePartyLocation,pan_current_y,f,Int32,0x71,1792,0,0,?
-SavePartyLocation,pan_finish_x,f,Int32,0x72,2304,0,0,?
-SavePartyLocation,pan_finish_y,f,Int32,0x73,1792,0,0,?
+SavePartyLocation,pan_current_x,f,Int32,0x70,0,0,0,int
+SavePartyLocation,pan_current_y,f,Int32,0x71,0,0,0,?
+SavePartyLocation,pan_finish_x,f,Int32,0x72,0,0,0,?
+SavePartyLocation,pan_finish_y,f,Int32,0x73,0,0,0,?
```
Ghabry: iirc I directly ripped the save defaults out of RPG_RT.exe, checking this later

CherryDT:
stop_count is counted towards max_stop_count until the next move is possible, this is used to wait during the "wait" move event, for movement frequency <8 and such. anim_count is for timing the stepping animation. These apply not only to the party but also the events and vehicles (the chunk id is below 100 so it's still in the shared range of SavePartyLocation, SaveEvent and SaveVehicleLocation)

src/rpg_fixup.cpp
fmatthew5876: A lambda where you can say `auto`. These 2 lambdas are identical except for the type of the parameter.

generator/csv/fields.csv
fmatthew5876: Those might be the correct defaults, but in Player we set them to 0. Then they get written back out to the LSD.

I don't know what values these are in `RPG_RT.exe`, but they are defaulted and not written out.

It's possible your defaults were correct and we are interpreting them differently in Player.

CherryDT: The defaults are indeed 9*256 and 7*256 (center - as the panning is stored in 1/16 pixels)

fmatthew5876: Thanks, I'll take a look at how this is used in Player and come up with a better solution here.

CherryDT:
Unrelated, but I didn't succeed connecting to the IRC with any of the clients which I tried: Something is wrong, Discord seems to be disconnected from IRC since today or so!

fmatthew5876:
Looks like we don't support `SaveInventory::steps` at all in Player. Will make a separate PR for that.

For `frame_count`, does it just count up to infinity? I imagine this number can get very large fast. Does it overflow and wrap at some point?

CherryDT:
`frame_count` just counts up indefinitely. (It would go haywire after 414 days of continuous playing... so not really important to consider.) It is used only for two things really: One is for drawing animated map tiles (their animation frame is based on it - *probably* you can ignore that because except for the very first map it's pretty unpredictable anyway what value the frame count will have, so it is unlikely that a game depends on that graphically), and for the animation of the character graphics and the animated arrows in the shop when buying equipment (I guess you don't need to care about that).

No, `panorama_data` is its own structure, should probably be called `SavePanorama`, with X and Y in chunks 1 and 2.

fmatthew5876:
This `panorama_data` feature doesn't appear to be implemented in rm2k steam. I created a 20x15 map and set the background to scroll horizontally with speed 8. When I save the game, `RPG_RT` always writes an empty chunk at 0x70. When I load the game, the background is reset to it's default position.

What this feature removed from some later versions of rm2k?

generator/csv/fields.csv
fmatthew5876: This change was removed.

CherryDT:
It was implemented in RM2k3E v1.12, not RM2kE.

SaveSystem::screen: already answered Ghabry
SavePartyLocation::stop_count: I answered it above

Pictures: It shouldn't matter how many pictures exist in there as long as there aren't too many (although I'm not sure about that either - needs a test whether it crashes when the LSD has _too many_ pictures) - should be fine to grow the picture array dynamically, like the switch/variables array work. And you cannot detect the amount of supported pictures (without analysing RPG_RT.exe, that is).

fmatthew5876:
jenkins: test this please :heart:

fmatthew5876:
The overall LSD work is ongoing, but this set of commits up to this comment are ready to go if approved ( same for player PR).

src/generated/rpg_savesystem.h
```diff
@@ -26,6 +26,28 @@
 namespace RPG {
 	class SaveSystem {
 	public:
+		enum ScreenRpgRt {
```
fdelapena: Because they seem to be scene names, maybe scene is a better and shorter name instead of screen.

generator/csv/fields.csv
```diff
@@ -1083,7 +1085,7 @@ Save,actors,f,Array<SaveActor>,0x6C,,1,0,array of RPG::SaveActor
 Save,inventory,f,SaveInventory,0x6D,,0,0,RPG::SaveInventory
 Save,targets,f,Array<SaveTarget>,0x6E,,1,0,array of RPG::SaveTarget
 Save,map_info,f,SaveMapInfo,0x6F,,0,0,RPG::SaveMapInfo
-Save,panorama_data,f,Int32,0x70,-1,0,0,Irrelevant; used in game to store panorama data but the object does not have members exported when saved. Mostly an empty object.
+Save,panorama,f,SavePanorama,0x70,,1,0,Used to store panorama position data. Not implemented in some versions of RPG_RT?
```
Ghabry: Could you replace "Not implemented in some versions of RPG_RT?" with e.g. "Used by RPG_RT 2k3 1.12 in other versions an empty object"

src/generated/rpg_savesystem.h
Ghabry: Also find this RpgRt pretty ugly, but no idea how to name it. Screen oder Scene are an option. (and maybe add a comment that this is RPG_RT only).

fmatthew5876: I renamed it to just "Screen". I agree with Fdelapena that Scene is a better name than Screen, but Screen seems to be the terminology used by `RPG_RT`. Also since we call our thing a scene, using the word screen here helps avoid confusion.

CherryDT: Actually, the RPG_RT terminology is "scene" as well. (also in RGSS in later versions)

fmatthew5876: Well I guess we need to bikeshed this thing then. Here are some options:

* SceneRPG_RT / scene_rpgrt
* LegacyScene / legacy_scene
* Scene / scene
* Scene / scene - and change Player code / enums to use this

Votes anyone?

Ghabry: We use Scene because RGSS calls it Scene ;).

> Scene / scene
My favourite. When you document this as "Only used by RPG_RT, not by EasyRPG Player" (which you already do) I think this is not confusing and fine.

> and change Player code / enums to use this
Another possibility but I'm against this because we have more scenes than RPG_RT which would mean adding custom scene entries to liblcf... not a fan of this.

generator/csv/fields.csv
```diff
@@ -705,7 +705,7 @@ SaveTitle,face3_name,f,String,0x19,,0,0,char[]: face filename
 SaveTitle,face3_id,f,Int32,0x1A,0,0,0,int: face id
 SaveTitle,face4_name,f,String,0x1B,,0,0,char[]: face filename
 SaveTitle,face4_id,f,Int32,0x1C,0,0,0,int: face id
-SaveSystem,screen,f,Int32,0x01,1,0,0,
+SaveSystem,screen,f,Enum<SaveSystem_Screen>,0x01,0,1,0,The current Screen (Scene) type for RPT_RT. Legacy field only used by RPG_RT and not by EasyRPG Player.
```
Ghabry: RPT_RT again btw

fmatthew5876:
Comments addressed.

src/rpg_fixup.cpp
```diff
@@ -124,13 +178,125 @@ void RPG::SaveSystem::Fixup() {
...
+	auto reset_bgm = [](RPG::Music& s) {
+		s = {};
+		s.name.clear();
+	};
+
+	if (graphics_name == system.system_name) {
+		graphics_name.clear();
+	}
+
+#if 0
```
Ghabry: The whole if 0 block can be removed

fmatthew5876:
Removed the if 0. Great catch btw.

Ghabry:
I think the rebase failed.

Ghabry:
nice solution with this "UnFixup" btw, my initital idea for this was to get completely rid of "Fixup" and handle this in Player Getters, but I prefer yours.

fmatthew5876:
Still need to call unfixup for event info. That requires knowing the active page for the event at the liblcf level. Haven't figured out the best solution there yet.

PR #279, opened 2018-11-07
Save data animation_type defaults should be 0

fmatthew5876:
even animation type is supposed to default to 0 as in `RPT_RT`.

If you save your game in `Player`, and then load the game with `RPG_RT` all of the event sprites including the player will be walking in place.


fmatthew5876:
`SaveActorInfo::status` has a very unique behavior.

When you start a new game, all actors in the party get their status arrays resized. All actors not in the party have their status vectors empty.

The behaves the same way if you edit the game and add more states. When loading a saved game with fewer states, the actors in the party are resized but the others unchanged.

When an actor is added to the party, their state vector is resized. Whenever a state is added, removed, or even queried in a conditional branch on any actor, including those not in the party, their status vector is resized.

Instead of all that complication, I just resize the status array on new game and load game for all actors.

Ghabry:
I never heard of this "walking in place" issue when loading a Player save in RPG\_RT o.O.
Will test first if this has an annoying side effect for old EasyRPG saves (all actors walk in place ^^')

fmatthew5876:
I'd like to know why loading the save game in easyrpg doesn't make events walk in place. Maybe another lsd bug.

PR #282, opened 2018-11-18
Implement structs inheritance with SaveMapEventBase

fmatthew5876:
This PR establishes a C++ inheritance relationship between player, vehicle, and map event save data.

This will be used for https://github.com/EasyRPG/Player/issues/1499. It also reduces duplication in our csv files.

Also fixes bugs where `sprite_transparent` and `route_through` chunks were missing for some of these structs.

party move speed initialization was moved to Setup(), since it now defaults to -1 from inheritance.


fmatthew5876:
If we are happy here, this can be merged without breaking master in Player.

fmatthew5876:
This has been rebased over the Jinja changes

PR #283, opened 2018-11-24
Fix SaveActor::row

fmatthew5876:
Changed it to an enum. The values I use are:
* 0 - front
* 1 - back

How did we come up with -1 meaning front, 1 back, and 0 being a meaningless default? Since the row chunk is not present in save data when defaulted, there doesn't seem anyway to know what the front value is in `RPG_RT` without looking at process memory, but I can't see why it would be -1.


Ghabry:
No idea how I came up with -1, guess I changed it until it loaded correctly from a RPG_RT save ;)

Issue #245, opened 2018-10-02
Add enums for event command parameters


fdelapena:
Currently Player uses dozens of [magic numbers for event command parameters](https://github.com/EasyRPG/Player/search?q=com.parameters). For more readable code, mostly in game interpreter and shared and better support for game editors, liblcf might provide some generated code from CSV with these enumerations.

jetrotal:
I did some research on this:

Ghabry Suggested checking R48's commands descriptions, from @20kdc repos:
https://github.com/20kdc/gabien-app-r48/blob/master/app/src/main/resources/assets/R2K/CommandsGI_0.txt
![image](https://github.com/EasyRPG/liblcf/assets/45118493/00031838-53b1-404f-b91e-10313e41f643)

I also found some interesting references from Destiny Patch sourcecode (thanks to Kotatsu Akira and @drxgb)
[Object_Command.zip](https://github.com/EasyRPG/liblcf/files/12514111/Object_Command.zip)

![image](https://github.com/EasyRPG/liblcf/assets/45118493/20529618-7651-4d3e-bd74-0d07bafa32c0)
![image](https://github.com/EasyRPG/liblcf/assets/45118493/299bb649-7581-4069-bc9e-9181f4d16170)

The Destiny Patch one also has lots of name suggestions for some parameters outputs, 
maybe making things friendlier for the creation of a scripting language.

20kdc:
just so you know, some commands are... complicated, and can't be directly dumped from r48's CMDB due to this
that said if you need a JSONification of what's available I can rig up something if asked

20kdc:
This isn't absolutely all data in the system (it gets increasingly complicated to do that going further into the details) but it should have an entry for every command.
[cmdb.json.zip](https://github.com/EasyRPG/liblcf/files/12515580/cmdb.json.zip)

20kdc:
Oh, right, I should probably note:

* `specialSchemaEssential` is a flag indicating that the parameters given aren't necessarily accurate and basically you can chuck them out a window, because the command does something really specific that can't be expressed with a simple array-of-parameters format.
* `nameRawUnlocalized` is written using S-expressions and custom stuff. Details are in the R48 development documentation, but I wouldn't really recommend trying to parse these unless you have to. They're basically inline stringifier logic.
* `name` is directly derived from `nameRawUnlocalized`, i.e. you can derive `name` from `nameRawUnlocalized` given the prerequisite support code.
[...]

PR #287, opened 2018-12-02
Add support for generated constants in RPG:: classes

fmatthew5876:
Towards fixing https://github.com/EasyRPG/Player/issues/1516

Also adds panning constants, so that the default pan values don't need to be duplicated between player and liblcf.

### DIFF
--- a/src/generated/rpg_savepartylocation.cpp
+++ b/src/generated/rpg_savepartylocation.cpp
@@ -1,0 +1,17 @@
+/* !!!! GENERATED FILE - DO NOT EDIT !!!!
+ * --------------------------------------
+ *
+ * This file is part of liblcf. Copyright (c) 2018 liblcf authors.
+ * https://github.com/EasyRPG/liblcf - https://easyrpg.org
+ *
+ * liblcf is Free/Libre Open Source Software, released under the MIT License.
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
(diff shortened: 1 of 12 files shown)


carstene1ns:
Thanks! This will also be very useful for #245.

Ghabry:
Lets make a quick style discussion:
I find Hungarian notation really ugly. Is that some common idiom in C++11 code to prefix constexpr-constants with a German "konstant"-k?

Ghabry:
Ah I thought it is k because c is the prefix for classes.

I dislike using e for enum but for the k we did a quick poll and the result is:

We don't **k**are

src/generated/rpg_savepartylocation.cpp
```diff
@@ -0,0 +1,18 @@
...
+ * liblcf is Free/Libre Open Source Software, released under the MIT License.
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
+// Headers
+#include "rpg_savepartylocation.h"
+constexpr int RPG::SavePartyLocation::kPanXDefault;
+constexpr int RPG::SavePartyLocation::kPanYDefault;
+
```
Ghabry: I would like an empty line after the include and these 3 empty lines at the end removed

fmatthew5876:
Fixed the whitespace. I also redefined the panning constants as X * 256

PR #288, opened 2018-12-02
BUGFIX: Don't generate tags for unsupported enums

fmatthew5876:
enum tags only works for enums which start from 0
and increase monotonically by 1. Until we have a better
solution, don't generate tags for enums where it would break.


fmatthew5876:
Something like this might be a better all around solution:

https://github.com/quicknir/wise_enum

https://www.youtube.com/watch?v=GCkhcT2oxCA

Ghabry:
Hm, this could've been noticed during the review.

Ironically you wrote in another PR that you wanted this feature especially for the EventCommand and Movement enum and now you remove half of it ;).

fmatthew5876:
wise enum seems to have a 256 value limit, but it sounds like they have a codegen script if you want more. The one thing I like about the macro approach is that generates a switch statement which leaves the optimal implementation in the hands of the compiler instead of the programmer.

We could continue with the tags approach. We would need to support an offset for monotonic enums starting > 1, and a binary search (or perfect hash if you want to be crazy) style for unstructured enums.

PR #289, opened 2018-12-02
Parallelize reader_struct build

fmatthew5876:


Ghabry:
Maybe this also helps retropie (raspberry pi) users because compiling readerstruct runs out of memory and terminatens gcc :D

PR #290, opened 2018-12-02
Fix clang warnings about templates

### LINKED ISSUE #221: clang dislikes reader_struct and reader_flags
Lot's of these (122):
```
src/reader_struct.cpp:27:18: warning: instantiation of variable 'Struct<RPG::Actor>::fields' required here, but no definition is available [-Wundefined-var-template]
        for (int i = 0; fields[i] != NULL; i++)
                        ^
./src/reader_struct.h:512:25: note: forward declaration of template entity is here
        static const Field<S>* fields[];
                               ^
src/reader_struct.cpp:27:18: note: add an explicit instantiation declaration to suppress this warning if 'Struct<RPG::Actor>::fields' is explicitly instantiated in another translation unit
        for (int i = 0; fields[i] != NULL; i++)
                        ^
```

fmatthew5876:
Depends on #289 
Fixes #221 

MSVC chokes on this so I had to split it out. Maybe upgrading to VS2017 will fix it?

### DIFF
--- a/src/reader_struct_impl.h
+++ b/src/reader_struct_impl.h
@@ -271,3 +271,292 @@ void Struct<S>::BeginXml(std::vector<S>& obj, XmlReader& stream) {
 	stream.SetHandler(new StructVectorXmlHandler<S>(obj));
 }
 
+template <>
+char const* const Struct<RPG::Actor>::name;
+template <>
+Field<RPG::Actor> const* Struct<RPG::Actor>::fields[];
+
+template <>
+char const* const Struct<RPG::Animation>::name;
+template <>
+Field<RPG::Animation> const* Struct<RPG::Animation>::fields[];
+
+template <>
+char const* const Struct<RPG::AnimationCellData>::name;
+template <>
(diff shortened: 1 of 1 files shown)


carstene1ns:
IMO we could wait with this until our build agent is updated with vcpkg and MSVC 2017. Maybe is really fixed and does not break the build currently.

Ghabry:
@fmatthew5876 
Compiler is now latest VS2017, try again building under Windows.

fmatthew5876:
Looks like no.

Lets leave this alone for now. There are other ways to refactor the code that should be nicer than some big ugly ifdef. For example, probably changing `Struct::name()` to an inline function will fix it for all platforms.

Ghabry:
a search suggested to use "/permissive-" but I just tested and didn't work either.

carstene1ns:
What I dislike here is that this list is not autogenerated.

jenkins: test this please

carstene1ns:
Closing. Can be reopened when there is a proper solution to this.

PR #291, opened 2018-12-02
Make all liblcf structs comparable

fmatthew5876:
* Simplifies and streamlines comparison logic
* Removes meta function hacks around comparison in `reader_struct`
* Adds additional utility for users of liblcf


fmatthew5876:
@Ghabry This one will require us to re-test copying for LDB, LMT, LMU, and LSD. I'll check on my side today but please if you could also run your test suite again to sanity check me.

PR #292, opened 2018-12-06
Composite Characters Fix

SmiVan:
[#1530](https://github.com/EasyRPG/Player/issues/1530) and probably [#1383](https://github.com/EasyRPG/Player/issues/1383) and [#839](https://github.com/EasyRPG/Player/issues/839) in Player, by implementing normalization in place of lower-casing per @Ghabry's patch in [#839](https://github.com/EasyRPG/Player/issues/839).

Requires more thorough testing, but is confirmed to work against the Yume2kki example in [#1530](https://github.com/EasyRPG/Player/issues/1530).

***
Also needs to be tested that it doesn't break everything on [non-Mac platforms which use NFC as opposed to NFD](https://gist.github.com/JamesChevalier/8448512).

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -13,8 +13,8 @@
 #ifdef LCF_SUPPORT_ICU
 #   include <unicode/ucsdet.h>
 #   include <unicode/ucnv.h>
-#	include <unicode/normalizer2.h>
-#	include <unicode/unistr.h>
+#   include <unicode/normalizer2.h>
+#   include <unicode/unistr.h>
 #else
 #   ifdef _MSC_VER
 #		error MSVC builds require ICU
@@ -387,7 +387,6 @@ std::string ReaderUtil::Normalize(const std::string &str) {
 	f.toUTF8String(res);
 	return res;
 #else
-	// LAME
 	std::string result = str;
 	std::transform(result.begin(), result.end(), result.begin(), tolower);
 	return result;
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -379,12 +379,16 @@ std::string ReaderUtil::Recode(const std::string& str_to_encode,
 
 std::string ReaderUtil::Normalize(const std::string &str) {
 #ifdef LCF_SUPPORT_ICU
-	icu::UnicodeString& uni = icu::UnicodeString(str.c_str()).toLower();
+	icu::UnicodeString uni = icu::UnicodeString(str.c_str()).toLower();
 	UErrorCode err = U_ZERO_ERROR;
+	std::string res;
 	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
 	icu::UnicodeString f = norm->normalize(uni, err);
-	std::string res;
-	f.toUTF8String(res);
+	if (U_FAILURE(err)) {
+		uni.toUTF8String(res);
+	} else {
+		f.toUTF8String(res);
+	}
 	return res;
 #else
 	std::string result = str;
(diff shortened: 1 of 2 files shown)


src/reader_util.cpp
```diff
@@ -13,6 +13,8 @@
 #ifdef LCF_SUPPORT_ICU
 #   include <unicode/ucsdet.h>
 #   include <unicode/ucnv.h>
+#	include <unicode/normalizer2.h>
```
Ghabry: whitespace issues

src/reader_util.cpp
```diff
@@ -374,3 +376,20 @@ std::string ReaderUtil::Recode(const std::string& str_to_encode,
...
+#ifdef LCF_SUPPORT_ICU
+	icu::UnicodeString& uni = icu::UnicodeString(str.c_str()).toLower();
+	UErrorCode err = U_ZERO_ERROR;
+	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
+	icu::UnicodeString f = norm->normalize(uni, err);
+	std::string res;
+	f.toUTF8String(res);
+	return res;
+#else
+	// LAME
```
Ghabry: please don't copy-paste sloppy stuff from my proposed patch, like this // LAME comment.

Anybody having documentation for iconv around? I assume iconv should also support this.

Ghabry:
I made more tests months ago concerning the decomposition, will check if I used NFKC or NFKD there.

It is extremely unlikely that the Normalization will break on any other platforms, the whole procedure is standardized, so for one input, even when in different forms, you will get the same NFKC output.

Look at Figure 6: http://unicode.org/reports/tr15/

The problem with the "K" form are some corner cases. E.g. when you have a file with name 2³ and a file 23 they will have after NFKC the same codepoints.
The same applies to a fi ligature which becomes f + i.
However these corner cases can't really happen when you convert from codepages to unicode, it could be a problem later with the editor when full unicode support is there, but I still consider this problem academic, we can fix it when it ever becomes a problem.

src/reader_util.cpp
fdelapena: https://stackoverflow.com/a/23159081

Ghabry: lol, okay nevermind, good that we have ICU.

fdelapena:
Jenkins: ok to test

src/reader_util.cpp
SmiVan: yep, I think I pasted in a tab by accident

src/reader_util.cpp
```diff
@@ -374,3 +376,19 @@ std::string ReaderUtil::Recode(const std::string& str_to_encode,
 	return result;
 #endif
 }
+
+std::string ReaderUtil::Normalize(const std::string &str) {
+#ifdef LCF_SUPPORT_ICU
+	icu::UnicodeString& uni = icu::UnicodeString(str.c_str()).toLower();
+	UErrorCode err = U_ZERO_ERROR;
+	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
+	icu::UnicodeString f = norm->normalize(uni, err);
```
carstene1ns: Since ICU will not touch the destination string in error case, maybe we should return the former string?
```c++
if (U_FAILURE(err))
```

Ghabry: added return of lowercase string in error case and removed the &-ref, that looked like a bug.

Ghabry:
Oh, this is even documented and not an implementation detail, great. Will remove that one error check before merging

PR #298, opened 2018-12-14
Small Maintenance

carstene1ns:


CMakeLists.txt
```diff
@@ -78,44 +96,125 @@ add_library(lcf
...
+	src/generated/rpg_variable.h
 )
 
+# IDE source grouping
+foreach(SG LDB LMT LMU LSD RPG)
+	string(TOLOWER ${SG} LSG)
+	source_group("Source Files\\generated\\${SG}" REGULAR_EXPRESSION "generated/${LSG}_.*\\.cpp")
+	source_group("Header Files\\generated\\${SG}" REGULAR_EXPRESSION "generated/${LSG}_.*\\.h")
+	source_group("Source Files\\${SG}" REGULAR_EXPRESSION "${LSG}_.*\\.cpp")
+	source_group("Header Files\\${SG}" REGULAR_EXPRESSION "${LSG}_.*\\.h")
```
Ghabry: Can you tell me any IDE (on Linux) where I can see the result?

CMakeLists.txt
```diff
@@ -78,44 +96,125 @@ add_library(lcf
...
+endforeach()
+
 # C++11 is required
 target_compile_features(lcf PUBLIC cxx_std_11)
```
Ghabry: while we are at maintanance could you change this to the CMake 3.7 compatible:

```
set_target_properties(lcf PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED ON
    CXX_EXTENSIONS OFF
)
```

CMakeLists.txt
carstene1ns: e.g. Eclipse:
`$ cmake . -DCMAKE_BUILD_TYPE=Release -G "Eclipse CDT4 - Ninja" -DCMAKE_ECLIPSE_VERSION=4.9`

bloat

CMakeLists.txt
carstene1ns: I have instead used the older default value for this, as the tests would not get the required `-std=c++11` parameter otherwise without specifying it again.
Not a problem on current compilers, as their default may be c++14 already.
https://cmake.org/cmake/help/v3.8/variable/CMAKE_CXX_STANDARD.html

PR #301, opened 2019-01-01
Add LCF mime types and install them

carstene1ns:


Ghabry:
Oh, you already force-pushed away the issue I noticed yesterday.

The other problem (or discussion topic) I see is: Is it intended to alter the system mime database when cross-compiling? Imo a cross-compile should not modify the host-system it is compiled on. In CMake this is detectable via CMAKE_CROSSCOMPILING (is broken on OSX but not relevant here).

carstene1ns:
Added a commit that fixes this.

Issue #174, opened 2016-03-15
SaveMapEvent pending has unknown data type


Ghabry:
The current assumption is BOOL but we got send a Ib savegame (created with RPG_RT, we don't write this field) that has a size of 4 and value 257. So this is not a compressed integer. Possible are Vector<Uint8> or Int (but if the field can be < 4 this will fail again).

[Save01.zip](https://github.com/EasyRPG/liblcf/files/174770/Save01.zip)

Ghabry:
@CherryDT while you are around ;)

Can you tell us the purpose and format of chunk 0x67 for saved map events? 
Our guess was "This event will run after the current running event stops running." but we had a bool-flag for it and then got a save sent that had a non-bool value in this chunk.

CherryDT:
That's whether the event was started by action button. It's what's used for the corresponding conditional branch option.

(Btw it's so weird you guys use hex for these chunks... They are actually decimal and have a numbering scheme...)

CherryDT:
@Ghabry I know this doesn't belong here but GitHub has no private messaging anymore... Is your g...k@web.de email still active?

Ghabry:
Yes it is still active.
My new address is gabriel at this domain: https://mastergk.de/

Ghabry:
>That's whether the event was started by action button. It's what's used for the corresponding conditional branch option.

Really? We use 0x0D in SaveEventCommands for this

CherryDT:
That's also correct. They are stored in both the event and the corresponding scripter class, hence why they exist twice in the save file too.

Ghabry:
interesting >.> and that the value was 257 in one savegame, sounds not very bool-ish :D

Ghabry:
The chunk has a proper name now and the parser can handle corrupt chunks well and there is no way for us to support data encoded in higher bytes for RPG_RT patches. Therefore I consider this as fixed.

PR #310, opened 2019-02-21
Fix MapEvent  waiting and trigger chunks

fmatthew5876:
This is the flag used to say an event is waiting for
foreground execution.

As described by Cherry in https://github.com/EasyRPG/Player/issues/1247

For `waiting_execution` chunk, simply create 2 autostart events with
```
OpenSaveMenu
```

The second event will have the flag set, while the first one (currently executing) does not.

For the `trigger_by_decision_key` chunk. Change the above events to trigger type `action`, `touch`, and/or `collision` and talk to them with the decision key. The chunk will be set in the save data.


Ghabry:
For consistency with Player code I would prefer the past tense ``triggered_by_decision_key``

fmatthew5876:
Changed it to "triggerred"

carstene1ns:
rel. #174

PR #311, opened 2019-02-21
Rename event interpreter runtime state chunks

fmatthew5876:
Depends on #310 

This renames a bunch of event interpreter chunks and structs to better reflect their actual purpose. #310 is pretty much a no brainer, but this one will require a PR in player (https://github.com/EasyRPG/Player/pull/1643) and also maybe you guys have opinions on the names?

* `SaveEventRunFrame` - There is usually only one of these, except when you do `Call Event` we push more of them, hence "RunFrame" and `SaveEventRunState::stack`
* foreground and background states accurately named.


Ghabry:
Finding an agreement here will be a problem as I have completely different opinions on the naming.
The only one I agree with is "stack".

Imo the agreed terminology by now is, at least this is what I use everywhere in conversations: "Main interpreter" for that one interpreter where all events except parallel runs and "Parallel interpreter" for all parallel events. So for me these are main and parallel event states.

Foreground <-> Main
Background <-> Parallel

Don't think the "Run" (one chunk is called wait_execution btw so this should be "Execution" then?) is needed, how about "SaveEventState"? I agree that EventData is misleading I'm also always fairly confused by this.

I needed a while to understand why it is "Run Frame" but you call it frame because it is a stackframe.
Imo this is a highly technical term. The main purpose of that class is what it is called now: It hold EvenCommands, therefore I like the old name.

Anybody else with opinions?

fmatthew5876:
> Finding an agreement here will be a problem as I have completely different opinions on the naming.
> The only one I agree with is "stack".
> 
> Imo the agreed terminology by now is, at least this is what I use everywhere in conversations: "Main interpreter" for that one interpreter where all events except parallel runs and "Parallel interpreter" for all parallel events. So for me these are main and parallel event states.
> 
> Foreground <-> Main
> Background <-> Parallel

To me foreground and background is a bit more obvious. I think of it like jobs in the bash shell. But if we already use the main/parallel terminology I'm fine switching to that.

Would like to know what others think.

> 
> Don't think the "Run" (one chunk is called wait_execution btw so this should be "Execution" then?) is needed, how about "SaveEventState"? I agree that EventData is misleading I'm also always fairly confused by this.

I would put "Run" or "Exec" or something like this in the name. Basically we should make it obvious that these chunks are for the runtime or execution state of the interpreter. The purpose of these chunks wasn't at all apparent to me until I really started digging into the interpreter.

I completely overlooked hidden in plain sight `Save::events` chunk for example.

Also the fact that we had "commands" repeated as a chunk name many times for different things was super confusing, especially when I was looking at them in `LCF_DEBUG_TRACE`.

> 
> I needed a while to understand why it is "Run Frame" but you call it frame because it is a stackframe.
> Imo this is a highly technical term. The main purpose of that class is what it is called now: It hold EvenCommands, therefore I like the old name.

It stores a bit more than just the commands. It also stores the index to the currently executing command along with some other running state. Also, this should be differentiated from chunks like `CommonEvent::event_commands`.

This is why I like the "stack frame" analogy, because that's really describing what it is.

> 
> Anybody else with opinions?

Ghabry:
Jenkins: test this please

fmatthew5876:
We've got to come to some consensus on this. I want to start using these chunks in intepreter refactors.

I will stick to my guns in that I much prefer foreground and background over main and parallel.

For runstate and runframe, I could be convinced otherwise. Only not something so meaningless as `EventData` and `EventCommands`.

@carstene1ns and @fdelapena 

Your opinions?

Ghabry:
The only thing I prefer is "Exec" instead of "Run" (or the annoying long Execution which is more correct from a dictionary point of view) 
For everything else c1 may have the final word. I let the vote decide.

Somehow I didn't notice last time that this only affects Save chunks, so I actually care less about the renaming because these chunks won't be really ever used outside of Player (and won't see them often even if I dislike the naming).

fmatthew5876:
I rebased this and changed `Run` to `Exec`.

Once this PR is approved I will update https://github.com/EasyRPG/Player/pull/1643 to match

fdelapena:
Jenkins: test this please

carstene1ns:
I actually have no preference here, just used to parallel.
shrug

Ghabry:
Okay, does this mean that we finally came to an agreement here?

fmatthew5876:
rebased to master

Ghabry:
Jenkins: Test this please

fmatthew5876:
I was thinking about this again.

How about foreground and parallel? Instead of foreground and background.

That seems to fit better into rpg maker vocabulary to me.

Ghabry:
That's even better because this is half of what I suggested initially ;)

fmatthew5876:
Alright, I've rebased this and the player PR using the new names. I think we have agreement now?

carstene1ns:
Just found out here have been added files not part of the build system (and therefore also not tarball):
```
src/generated/lsd_saveeventexecdata.cpp
src/generated/rpg_saveeventexecdata.h
src/generated/rpg_saveeventrunframe.h
src/generated/rpg_saveeventrunstate.h
```
What is the deal here? Messed up rebase?

PR #314, opened 2019-03-02
Handle error in Unicode normalization

Ghabry:
that's some problem with the data file but can't tell you why.

There is a way to debug print the datafile stuff with UDATA_DEBUG but this gives a build error xD

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -379,6 +379,15 @@ std::string ReaderUtil::Normalize(const std::string &str) {
 	UErrorCode err = U_ZERO_ERROR;
 	std::string res;
 	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
+	if (U_FAILURE(err)) {
+		static bool err_reported = false;
+		if (!err_reported) {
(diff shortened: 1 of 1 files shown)


Ghabry:
hm still not working because there seems to be a huge problem with the ICU data file:

liblcf:  ucnv_open() error for source encoding "ibm-5348_P100-1997": U_FILE_ACCESS_ERROR

Ghabry:
well, this is not needed for emscripten anymore but that error check is still useful because failing to load normalisation from ICU will return a nullptr which is dereferenced one line later.

src/reader_util.cpp
```diff
@@ -379,6 +379,15 @@ std::string ReaderUtil::Normalize(const std::string &str) {
 	UErrorCode err = U_ZERO_ERROR;
 	std::string res;
 	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
+	if (!norm || U_FAILURE(err)) {
```
carstene1ns: I do not think the `nullptr` check is necessary here, since ICU should always set the error code.

PR #316, opened 2019-03-08
Add EventPage::Frame_middle2

fmatthew5876:


carstene1ns:
I do not quite like `middle2`, maybe is better to rename `middle` to `middle_default` or something instead.

fmatthew5876:
> 
> 
> I do not quite like `middle2`, maybe is better to rename `middle` to `middle_default` or something instead.

I will disagree with you on this one. I think `middle_default` confuses the issue. There isn't a default value kind of situation.

There are really 2 middle frames, so I chose `middle` and `middle2`. If anything, I'd prefer 1 stays as `middle` because that's the most common middle value and `middle` is the most obvious name someone would reach for in absence of other knowledge. 

3 should have the more complicated name. I like 'middle2` as its concise to the point "this is the 2nd middle frame", but I'd be open to changing it if someone has a better idea.

@Ghabry?
@fdelapena ?

carstene1ns:
That is ok, my idea just came from the fact that we reset to middle frame at multiple places in Player, so being some kind of default.

Ghabry:
Don't think this is solvable in a nice way.
middle could be renamed to default but then this is confusing when the 2nd option becomes middle.

I would keep middle as is because this "middle2" is really an exception that won't occur often in the code.

carstene1ns:
good. merging then.

PR #317, opened 2019-03-11
MorphOS build fix

BSzili:
On other platforms stdlib.h gets included by other headers, but not on MorphOS, where atoi doesn't get declared, so I added an explicit include in encoder.cpp.


carstene1ns:
jenkins: ok to test.

PR #323, opened 2019-03-25
Rename 2k3 Skill/Item state_effect -> reverse_state_effect

fmatthew5876:
The original name was not descriptive and is almost the same
as state_effects, which is the state array.


Ghabry:
This will also need a small PR to Player because "state_effect" is already used 3 times.

The new name is good because the value is true when the default state effect is inverted.

Issue #155, opened 2015-09-20
Broken reading of a Save of The Sacred Tears TRUE


Zegeri:
This save file shared by @carstene1ns for EasyRPG/Player#579 is incorrectly parsed:
http://f4ke.de/dev/easyrpg/TSTT/Save03.lsd

When run with _DEBUG, it'll warn about multiple skipped chunks and it'll crash because of running out of bytes. What breaks liblcf is a keyinput_variable (0x16) at 0x3846, inside the SaveEventData of the 8th CommonEvent.

```
16 01 8C 17 01 01
```
should be parsed as:
```
keyinput_variable (size 01) (value 8c)
keyinput_all_directions (size 01) (value 01)
```

But it's actually parsed as:
```
keyinput_variable (size 01) (value 8c 17)
[crap]
```

keyinput_variable saves the variable used by the last KeyInputProc. Here's the funny part. In this case, the last KeyInputProc saves the input value in the variable 908. Now: 908 = 0x03**8C**. So keyinput_variable seems to be actually saving the last byte, not the whole number.

Ghabry:
I can't even get the KeyInput in a simple test case to work.

Have an Auto Start Common Event that reads Left/Right and then Waits 1.0 second and a map event that changes its look based on the button. The look never changes :(

Zegeri:
That setup you've described works on my end. Try running it in TestPlay mode and looking at the table of variables during that second.

Ghabry:
Yeah the problem I encountered was a different bug

CherryDT:
This is a bug I knew about for a long time but not the cause. Turns RM incorrectly saves this variable as boolean (which internally just copies 1 byte) instead of a var-len integer, probably a copy-paste error from the surrounding booleans.

Unfortunately this means it's hard to fix for me because if I would simply change this into an integer, then RPG_RT would encounter the same problem you had for certain saves and crash with a "stream read error" for bytes with the highest bit set. I'll see, maybe I'll save in a new chunk and load from both so that compatibility (at least for variables <128) is still okay.

PR #324, opened 2019-04-06
Interpreter state chunk fixes

fmatthew5876:
Saves space and removes ugly code from https://github.com/EasyRPG/Player/pull/1739

### DIFF
--- a/generator/csv/fields.csv
+++ b/generator/csv/fields.csv
@@ -908,7 +908,7 @@ SaveTarget,map_x,f,Int32,0x02,0,0,0,int
 SaveTarget,map_y,f,Int32,0x03,0,0,0,int
 SaveTarget,switch_on,f,Boolean,0x04,False,0,0,bool
 SaveTarget,switch_id,f,Ref<Switch>,0x05,0,0,0,int
-SaveEventExecFrame,commands_size,f,Int32,0x01,0,0,0,int
+SaveEventExecFrame,commands,t,EventCommand,0x01,0,0,0,int
 SaveEventExecFrame,commands,f,Vector<EventCommand>,0x02,,1,0,event command list
 SaveEventExecFrame,current_command,f,Int32,0x0B,0,0,0,int
 SaveEventExecFrame,event_id,f,Int32,0x0C,0,0,0,0 if it's common event or in other map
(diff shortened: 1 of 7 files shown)


generator/csv/fields.csv
```diff
@@ -936,7 +936,7 @@ SaveEventExecState,keyinput_left,f,Boolean,0x24,False,0,0,
 SaveEventExecState,keyinput_right,f,Boolean,0x25,False,0,0,
 SaveEventExecState,keyinput_up,f,Boolean,0x26,False,0,0,
 SaveEventExecState,keyinput_timed,f,Boolean,0x29,False,0,0,
-SaveEventExecState,unused_wait_for_key_or_enter,f,Boolean,0x2a,False,0,0,If enabled; an event waits for either the confirmation key to be pressed or one of the keys defined by KeyInputProc before continuing. This flag seems to be unused though since it is never written to (keyinput_wait is used instead).
```
fmatthew5876: Whats up with this comment? I tested with rm2k3e and it definately uses this chunk when you do a "Wait for Key Input" command. That is the wait command variant that waits for input, not a key input proc.

generator/csv/fields.csv
```diff
@@ -936,7 +936,7 @@ SaveEventExecState,keyinput_left,f,Boolean,0x24,False,0,0,
...
+SaveEventExecState,wait_key_enter,f,Boolean,0x2a,False,0,0,Used for a wait command "Wait For Key Input" rm2k3 feature to wait for decsion key press.
```
carstene1ns: > decision

fmatthew5876: Fixed, thanks!

fmatthew5876:
I took out the `keyinput_variable` fix from here. It breaks #155 and https://github.com/EasyRPG/Player/issues/1674

We need a better solution for that, but we should consider it separately

Ghabry:
This keyinput variable naming is really ugly :/
Different devs for 2k Value! and 2k3.

How does 2k before Value! (1.50) store key input?

fmatthew5876:
It's horrible I agree but I don't know a better solution. With the old names I had to add comments in my PR explaining why I was assigning direction to shift. With the new ugly names, no comments needed.

Old rm2k uses only all directions, decision, and cancel chunks.

PR #325, opened 2019-04-06
Rename 2k / 2k3 input chunks

fmatthew5876:
Depends on #324


fmatthew5876:
This actually doesn't seem to controversial so I will just merged to #324

PR #330, opened 2019-06-03
Fix SaveActor Flags

fmatthew5876:
Fix: https://github.com/EasyRPG/Player/issues/1789

It's not entirely clear to me whether this logic belongs in liblcf or Player. It seems that the fact that we use the `SaveActor` struct directly in Player (vs copying the data out to another structure), and therefore needing this adjustment is an implementation detail of Player.

Thoughts?


Ghabry:
Hm, I guess "Dual Wield", etc. are only saved in 2k3 because they can be overwritten there via a class change?

I have a slight feeling that the Player should be responsible for ignoring the fields but I havn't fully decided yet.

fmatthew5876:
Moved the fix to https://github.com/EasyRPG/Player/pull/1794

This one now just marks these flags as 2k3 only. This has no actual effect, other than documenting the fact that they are 2ke flags using the code.

PR #331, opened 2019-06-06
Change items_size -> item_ids count field

fmatthew5876:
https://github.com/EasyRPG/Player/pull/1798

### DIFF
--- a/src/generated/rpg_saveinventory.h
+++ b/src/generated/rpg_saveinventory.h
@@ -49,7 +49,6 @@ namespace RPG {
 		return l.party_size == r.party_size
 		&& l.party == r.party
 		&& l.item_ids == r.item_ids
-		&& l.item_ids == r.item_ids
 		&& l.item_counts == r.item_counts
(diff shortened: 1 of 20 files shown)


src/generated/rpg_saveinventory.h
```diff
@@ -49,7 +48,7 @@ namespace RPG {
 	inline bool operator==(const SaveInventory& l, const SaveInventory& r) {
 		return l.party_size == r.party_size
 		&& l.party == r.party
-		&& l.items_size == r.items_size
+		&& l.item_ids == r.item_ids
```
Ghabry: fdelapena mentioned in chat that this line is duplicated. Have to admit that I never noticed this before in the other PRs, is this happening with all the "operator==" that have size fields?
Just a cosmetic issue, won't affect the result.

fmatthew5876: Made an issue for this. Needs to be fixed in the comparison generation script.

fmatthew5876:
Added the fix for size fields.

PR #337, opened 2019-07-20
Remove SaveInventory::party_size

fmatthew5876:
Depends on: https://github.com/EasyRPG/Player/pull/1850

### DIFF
--- a/src/generated/lsd_saveinventory.cpp
+++ b/src/generated/lsd_saveinventory.cpp
@@ -21,11 +21,10 @@ char const* const Struct<RPG::SaveInventory>::name = "SaveInventory";
 
 template <>
 Field<RPG::SaveInventory> const* Struct<RPG::SaveInventory>::fields[] = {
-	new TypedField<RPG::SaveInventory, int32_t>(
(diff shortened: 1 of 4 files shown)


src/generated/lsd_saveinventory.cpp
```diff
@@ -21,11 +21,10 @@ char const* const Struct<RPG::SaveInventory>::name = "SaveInventory";
 
 template <>
 Field<RPG::SaveInventory> const* Struct<RPG::SaveInventory>::fields[] = {
-	new TypedField<RPG::SaveInventory, int32_t>(
-		&RPG::SaveInventory::party_size,
+	new CountField<RPG::SaveInventory, int16_t>(
```
Ghabry: I can't remember how this worked:
Are both int32_t and int16_t written out as BER integer or does this change the written amount from 2 to 4 byte?

fmatthew5876: I forget why, but it has to be `Count<int16_t>` because the underlying values are `int16_t`

The actual size field should still be 32 bit, not that it matters here as the valid values are only 0-4.

PR #345, opened 2019-11-18
Crash Fix: SaveEventExecFrame::commands size field always present if default

fmatthew5876:
Fixes "List index has exceeded its valid range (0)" crash
when loading Player saves in RPG_RT.


Ghabry:
I approve this

PR #351, opened 2019-12-21
CMake: Prevent conflicts in the test runner

Ghabry:
To make this better extensible "check" is now not touched anymore (just created when missing).

test_runner target is changed to test_runner_lcf and check_lcf attaches itself to check as a dependency.

the dependency tree is now "check <- check_lcf <- test_runner_lcf"


Ghabry:
even without this PR the liblcf-player-test combination already works btw, so this PR just makes extending the check target cleaner.

PR #355, opened 2019-12-30
Rename grid_location to field_depth and add enum

fmatthew5876:
This chunk is the one labeled "Field Depth" in the terrain editor. If you customize the grid, it becomes deselected and goes to 0.

Shallow is 1 and deep is 2. This is also selectible from EnemyEncounter command.


carstene1ns:
:question:

fmatthew5876:
I'm not yet sure what the naming scheme should be here. Will rethink this.

There is already `RPG::System::BattleFormation` for the enum value.

PR #358, opened 2020-03-03
Fix size fields being skipped while parsing

Ghabry:
Does not impact compatibility because we ignore the size fields, was just super annoying on Windows because it spams the slow terminal :)


CMakeLists.txt
```diff
@@ -232,6 +232,10 @@ target_include_directories(
 	${CMAKE_CURRENT_SOURCE_DIR}/src/generated
 )
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+	target_compile_definitions(lcf PUBLIC _DEBUG=1)
```
fmatthew5876: Can we call this something like `LCF_DEBUG`?

Ghabry: okay.
I'm currently thinking about making the whole skip debug printout always on even in Release as this indicated a problem in our parser (unknown chunk).

Ghabry:
Closing this in favor of #359
Will open a new PR related to "Skip" instead when ready.

PR #359, opened 2020-03-10
BIG REGRESSION - Fix sized fields being written twice bug

fmatthew5876:
Any chunks with size or count fields were being written twice.

Was introduced in september by this commit

https://github.com/EasyRPG/liblcf/commit/1b4fcf507451fe22798a1288fa25cdf46647dc01


fmatthew5876:
@Ghabry maybe this is actually the source of those error messages you're seeing in #358 ?

Ghabry:
@fmatthew5876 
Ugh how I could I miss that the field names are twice in the list when I fixed this :/

PR #360, opened 2020-04-03
Refactor Enums

fmatthew5876:
WIP

Would like an initial review and agreement on the approach before making the change PR to Player.

TBD:
- [ ] How to ensure enum tags enum values are sorted in disjoint case at compile time and not lose constexpr. Or do we sort them at compile time?
- [ ] Ensure compile times are not negatively affected by too much meta programming
- [ ] Template specialization requires you to change namespace scopes which is painful. Is there a better way?


Ghabry:
Can you elaborate a bit the motivation/use case behind this?

EnumTraits looks like some "reflection" for enums (knows how many values it has).

What is the EnumTable?
Why ZStringView? (and what is the Z)

Ghabry:
And I recommend that you create some kind of convertor for Player. Last time I did such an complex renaming I wrote a python script that invoked make in a loop, parsed the clang error message, autofixed the line and reenvoked make until it was stuck in an error loop or built with success.

Avoids lots of rebase hell.

fmatthew5876:
It's another attempt to solve an unsolved problem in C++, how to nicely generate data structures for enums, the main use case being enum <-> string.

The goals are:
* Provide a better enum interface than the one we have now
* Cleanup our enums to use modern C++
* Support both monotonic and disjoint enum to string tables (EventCommand::Code)

This approach has the following classes.

* EnumTraits - Provides reflection for enums
* EnumSet - An efficient set type for enum values
* EnumTable - An efficient map type for enum -> type T. Right now this is just the implementation of EnumTags, but it could be used more generally.
* EnumTags - Wraps EnumTable<Enum,ZStringView> and provides global access to it.
* ZStringView - To implement EnumTags I needed to add this. For EnumTags we want to be efficient and capture string literals as `const char*`. The problem is `operator==` compares pointers so we have to wrap them in a type. It's `ZStringView` because unlike `string_view` in C++17, this one is null terminated (Z for zero)

I'm not even sure how much I like this so far. There is quite a bit of repetition in how you define these things. However given C++ lack of reflection, some amount of repetition is impossible to avoid unfortunately :/

fmatthew5876:
Taking this down while I rethink the approach.

Issue #342, opened 2019-09-07
Namespace and organize lcf headers


fmatthew5876:
We need to reorganize the code before we can declare the liblcf API stable.

- [ ] Make all users `#include <liblcf/foo.H>` to avoid header conflicts.
- [ ] Put all `liblcf` code in an `lcf::` namespace.

Ghabry:
- [ ] Get rid of everything in "Data::", global variables are ugly

fmatthew5876:
As a first step, we could move `Data::` to player.

There are some parts of the liblcf parser that depends on grabbing stuff from `Data::`. That has to be refactored.

fmatthew5876:
I'd like to propose we do this header file and namespace change sometime when we cut 6.2 release.

The main thing blocking me from doing it now is rebase hell. I wouldn't want to make a namespace PR here and in Player and have it stay for a long time and cause lots of rebase conflicts with other PRs.

Seems like right before a release when we have a rush to get everything in would be a good time to cut over.

Ghabry:
moving this to 0.7.0 for now

carstene1ns:
> Seems like right before a release when we have a rush to get everything in would be a good time to cut over.

*Yeah, because nobody makes mistakes when in a rush.* :roll_eyes: This is to be done right after a release, not before.

fmatthew5876:
Ok sure, but maybe we could try to coordinate it, like right after the 6.2 release so it doesn't stay pending too long? This one is a rebase nightmare.

PR #361, opened 2020-04-03
Namespace and reorganize liblcf

fmatthew5876:
Depends on: https://github.com/EasyRPG/Player/pull/2099

Fix: #342 
Fix: #362 

This PR does the following:
- [x] Put all liblcf symbols in `::lcf` namespace
- [x] Put all rpg symbols in `::lcf::rpg` namespace
- [x] Put all public header files in `src/lcf/` or `src/generated/lcf` subdirectory. They now must be included like `#include "lcf/data.h"`
- [x] Rename public rpg headers from `rpg_*.h` to `lcf/rpg/*.h`
- [x] Keep all private headers and source files in  `src/`
- [x] Install public headers to `$(includedir)/lcf` and change pkgconfig to use `$(includedir)` directly in cflags
- [x] Don't install private headers

Testing:
- [x] Test install works in cmake
- [x] Test install works in automake
- [x] Test that cmake and automake produce same installations
- [x] Build player against installation using automake / pkgconfig
- [x] Build player against installation using cmake files
- [x] Build player with local liblcf on linux
- [x] Build player with local liblcf on windows
- [x] Create a dist tarball and check that it compiles with autotools
- [x] Create a dist tarball and check that it compiles with cmake

Here is an example install from cmake
```
.
./lib
./lib/liblcf.so
./lib/liblcf.a
./lib/liblcf.so.0.0.0
./lib/cmake
./lib/cmake/liblcf
./lib/cmake/liblcf/FindICU.cmake
./lib/cmake/liblcf/liblcf-config.cmake
./lib/cmake/liblcf/FindEXPAT.cmake
./lib/liblcf.la
./lib/pkgconfig
./lib/pkgconfig/liblcf.pc
./lib/liblcf.so.0
./include
./include/lcf
./include/lcf/rpg_attribute.h
./include/lcf/rpg_save.h
./include/lcf/rpg_savepartylocation.h
./include/lcf/lsd_chunks.h
./include/lcf/rpg_savepicture.h
./include/lcf/rpg_start.h
./include/lcf/rpg_music.h
./include/lcf/rpg_saveeventexecstate.h
./include/lcf/rpg_treemap.h
./include/lcf/rpg_system.h
./include/lcf/rpg_savemapevent.h
./include/lcf/ldb_chunks.h
./include/lcf/rpg_battleranimationdata.h
./include/lcf/rpg_class.h
./include/lcf/rpg_switch.h
./include/lcf/rpg_event.h
./include/lcf/flag_set.h
./include/lcf/rpg_saveeventexecframe.h
./include/lcf/scope_guard.h
./include/lcf/rpg_movecommand.h
./include/lcf/rpg_equipment.h
./include/lcf/rpg_eventpagecondition.h
./include/lcf/data.h
./include/lcf/rpg_animation.h
./include/lcf/rpg_savetarget.h
./include/lcf/lmu_chunks.h
./include/lcf/rpg_variable.h
./include/lcf/rpg_parameters.h
./include/lcf/lcf_saveopt.h
./include/lcf/rpg_troopmember.h
./include/lcf/lsd_reader.h
./include/lcf/rpg_actor.h
./include/lcf/rpg_animationframe.h
./include/lcf/rpg_savemapeventbase.h
./include/lcf/rpg_map.h
./include/lcf/enum_tags.h
./include/lcf/rpg_savevehiclelocation.h
./include/lcf/rpg_skill.h
./include/lcf/rpg_itemanimation.h
./include/lcf/rpg_savepanorama.h
./include/lcf/rpg_savemapinfo.h
./include/lcf/rpg_eventpage.h
./include/lcf/encoder.h
./include/lcf/rpg_battlecommand.h
./include/lcf/rpg_trooppage.h
./include/lcf/rpg_commonevent.h
./include/lcf/rpg_enemy.h
./include/lcf/rpg_mapinfo.h
./include/lcf/rpg_trooppagecondition.h
./include/lcf/rpg_learning.h
./include/lcf/ldb_reader.h
./include/lcf/reader_xml.h
./include/lcf/rpg_battleranimation.h
./include/lcf/rpg_rect.h
./include/lcf/rpg_terrain.h
./include/lcf/rpg_savetitle.h
./include/lcf/lmt_reader.h
./include/lcf/reader_util.h
./include/lcf/ini.h
./include/lcf/rpg_battlecommands.h
./include/lcf/rpg_animationcelldata.h
./include/lcf/rpg_saveinventory.h
./include/lcf/rpg_animationtiming.h
./include/lcf/rpg_savecommonevent.h
./include/lcf/reader_lcf.h
./include/lcf/rpg_savesystem.h
./include/lcf/rpg_troop.h
./include/lcf/inireader.h
./include/lcf/rpg_moveroute.h
./include/lcf/rpg_item.h
./include/lcf/rpg_testbattler.h
./include/lcf/rpg_terms.h
./include/lcf/lmu_reader.h
./include/lcf/rpg_database.h
./include/lcf/lmt_chunks.h
./include/lcf/rpg_encounter.h
./include/lcf/rpg_saveeasyrpgdata.h
./include/lcf/writer_lcf.h
./include/lcf/rpg_enemyaction.h
./include/lcf/rpg_saveactor.h
./include/lcf/rpg_eventcommand.h
./include/lcf/rpg_battleranimationextension.h
./include/lcf/writer_xml.h
./include/lcf/rpg_state.h
./include/lcf/rpg_chipset.h
./include/lcf/rpg_sound.h
./include/lcf/command_codes.h
./include/lcf/config.h
./include/lcf/rpg_savescreen.h
./share
./share/mime
./share/mime/packages
./share/mime/packages/liblcf.xml
./share/mime/XMLnamespaces
./share/mime/generic-icons
./share/mime/treemagic
./share/mime/aliases
./share/mime/globs2
./share/mime/version
./share/mime/subclasses
./share/mime/application
./share/mime/application/x-ldb.xml
./share/mime/application/x-lmt.xml
./share/mime/application/x-lsd.xml
./share/mime/application/x-lmu.xml
./share/mime/mime.cache
./share/mime/globs
./share/mime/magic
./share/mime/types
./share/mime/icons
```


fmatthew5876:
I added a fix for #362 

Since `LCF_SUPPORT_XML` is needed in `reader_xml.h`. I changed cmake to also generated a `config.h` which defines this. 

Now whether liblcf was built with autotools or cmake, external applications linking against it can know whether or not it has ICU and/or XMl support.

fmatthew5876:
This has now been fully tested on windows and linux using the associated player PR.

Ghabry:
0.6.2 is near and this should land quite soon after.

---------------

Question is if there should be directly more logical folder grouping as e.g. the MSVC Generator through CMake does:

Unbenannt

Maybe also get rid of file prefix (also planned for Player when the Makefiles are gone):

``generated/lmt/lmt_encounter.cpp -> generated/lmt/encounter.cpp``

---------------

Namespacing as you asked about RPG namespace:

Currently we have 5 prefixes: ldb, lmu, lsd, lmt, rpg. But it seems that ldb, lsd, lmt and lmu are just for the template-foo in cpp files, so there are no changes for outside users there.

About RPG: Imo lcf::RPG looks strange because of the full lower-upper case. Should be lcf::rpg. Not sure about dropping rpg namespace completely because the namespace adds semantic information: "Everything in rpg is part of the data structure".

fmatthew5876:
> 0.6.2 is near and this should land quite soon after.
> 
> Question is if there should be directly more logical folder grouping as e.g. the MSVC Generator through CMake does:
> 
> Unbenannt
> 
> Maybe also get rid of file prefix (also planned for Player when the Makefiles are gone):
> 
> `generated/lmt/lmt_encounter.cpp -> generated/lmt/encounter.cpp`

Sure we can move this stuff around. I'm not convinced we'll want to have `lcf::ldb`, `lcf::lsd`, etc.. namespaces though. Seems like overkill maybe?
> 
> Namespacing as you asked about RPG namespace:
> 
> Currently we have 5 prefixes: ldb, lmu, lsd, lmt, rpg. But it seems that ldb, lsd, lmt and lmu are just for the template-foo in cpp files, so there are no changes for outside users there.
> 
> About RPG: Imo lcf::RPG looks strange because of the full lower-upper case. Should be lcf::rpg. Not sure about dropping rpg namespace completely because the namespace adds semantic information: "Everything in rpg is part of the data structure".

I agree about `lcf::rpg::` namespace. All the game "data" is nice to have in it's own space.


> 
> Do you already have plans about making enums -> enum class?

I'm on the fence about this one. We do a lot of enum <-> int conversions and enum class adds a lot of noise with needing to do `static_cast` everywhere and also fully scoping the enum name. If we go with enum classes, we may want to take the enums out from being members of their respective classes so we don't have to say `lcf::rpg::ClassName::EnumName::EnumValue`. Need to think about it more.

I will take another pass at this PR this weekend.

Ghabry:
> I'm not convinced we'll want to have `lcf::ldb`, `lcf::lsd`, etc.. namespaces though

Because this is not public API: No, namespacing is not needed.
Also too deep nesting looks ugly before C++17 because you can't do "namespace A::B::C {}"

> I'm on the fence about this one. We do a lot of enum <-> int conversions

Right the casting is very annoying with enum classes. Lets focus on the other parts first.

fmatthew5876:
Rebased, but I need to review and retest everything. Also add `lcf::rpg`

fmatthew5876:
I've made the following changes

- [x] Renamed `RPG::` to `lcf::rpg::`

- [x] Move headers `rpg_*.h` to `lcf/rpg/*.h`

- [x] Made `MoveCommand::Code` and `EventCommand::Code` enum classes

This gets rid of the ugly `struct Code { enum Index { ... }};` thing we have going now. The price of this is more static casts in player (you can see them in the player PR).

Since these really are codes and not meant to be used as numbers, and they already use this nested style, this one seemed ok to migrate to enum class.

The others we can consider in a follow up PR.


As for moving lmt, lmu, etc.. into their own directories I'm punting on this as it's not part of the public interface. If someone wants this badly they can make a later PR.

That is, unless we really want `lcf/lsd_reader.h` to be `lcf/lsd/reader.h`?

fmatthew5876:
I put the ldb, lmt, lmu, and lsd public headers in their own sub directories. The final installation looks much nicer that way.

Not happy with the cmake/automake hacks to make this work. Is there a better way?

Not sure about putting all the ldb, lmt, lmu, and lsd source files in subdirectories. Do we really want this? I don't have a strong opinion as it's not part of the public interface.

Ghabry:
I like it this way :+1: 

Most important part is to avoid another move-around in the public interface (requiring player updates). Internal interface can be moved around in a later PR.

The extra work for install really looks ugly. No idea yet if this is solvable in a better way, will think about it.

Ghabry:
One way to make CMake shorter is e.g. by using Regex for filtering as we do with the MSVC source groups and everything that was not hit by the filter is installed directly in lcf.

But this probably only hides the problem? No idea if this is better :)

```cmake
foreach(S ldb lmt lmu lsd rpg)
	set(SUBLIST ${LCF_HEADERS})
	list(FILTER SUBLIST INCLUDE REGEX "/${S}/")
	install(FILES ${SUBLIST} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/${S})
	list(REMOVE_ITEM LCF_HEADERS ${SUBLIST})
endforeach()
install(FILES ${LCF_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf)
```



```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 22af701..f59c7de 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,7 @@ target_sources(lcf
 	src/generated/rpg_savepartylocation.cpp
 	src/generated/rpg_state.cpp
 	src/generated/rpg_system.cpp
-	)
+)
 
 set(LCF_HEADERS
 	src/lcf/data.h
@@ -137,61 +137,46 @@ set(LCF_HEADERS
 	src/lcf/flag_set.h
 	src/lcf/ini.h
 	src/lcf/inireader.h
-	src/lcf/saveopt.h
+	src/lcf/ldb/reader.h
+	src/lcf/lmt/reader.h
+	src/lcf/lmu/reader.h
+	src/lcf/lsd/reader.h
 	src/lcf/reader_lcf.h
 	src/lcf/reader_util.h
 	src/lcf/reader_xml.h
+	src/lcf/saveopt.h
 	src/lcf/scope_guard.h
 	src/lcf/writer_lcf.h
 	src/lcf/writer_xml.h
-)
-
-set(LCF_LDB_HEADERS
-	src/lcf/ldb/reader.h
 	src/generated/lcf/ldb/chunks.h
-)
-
-set(LCF_LMT_HEADERS
-	src/lcf/lmt/reader.h
 	src/generated/lcf/lmt/chunks.h
-)
-
-set(LCF_LMU_HEADERS
-	src/lcf/lmu/reader.h
 	src/generated/lcf/lmu/chunks.h
-)
-
-set(LCF_LSD_HEADERS
-	src/lcf/lsd/reader.h
 	src/generated/lcf/lsd/chunks.h
-)
-
-set(LCF_RPG_HEADERS
 	src/generated/lcf/rpg/actor.h
-	src/generated/lcf/rpg/animation.h
 	src/generated/lcf/rpg/animationcelldata.h
 	src/generated/lcf/rpg/animationframe.h
+	src/generated/lcf/rpg/animation.h
 	src/generated/lcf/rpg/animationtiming.h
 	src/generated/lcf/rpg/attribute.h
 	src/generated/lcf/rpg/battlecommand.h
 	src/generated/lcf/rpg/battlecommands.h
-	src/generated/lcf/rpg/battleranimation.h
 	src/generated/lcf/rpg/battleranimationdata.h
 	src/generated/lcf/rpg/battleranimationextension.h
+	src/generated/lcf/rpg/battleranimation.h
 	src/generated/lcf/rpg/chipset.h
 	src/generated/lcf/rpg/class.h
 	src/generated/lcf/rpg/commonevent.h
 	src/generated/lcf/rpg/database.h
 	src/generated/lcf/rpg/encounter.h
-	src/generated/lcf/rpg/enemy.h
 	src/generated/lcf/rpg/enemyaction.h
+	src/generated/lcf/rpg/enemy.h
 	src/generated/lcf/rpg/equipment.h
-	src/generated/lcf/rpg/event.h
 	src/generated/lcf/rpg/eventcommand.h
-	src/generated/lcf/rpg/eventpage.h
+	src/generated/lcf/rpg/event.h
 	src/generated/lcf/rpg/eventpagecondition.h
-	src/generated/lcf/rpg/item.h
+	src/generated/lcf/rpg/eventpage.h
 	src/generated/lcf/rpg/itemanimation.h
+	src/generated/lcf/rpg/item.h
 	src/generated/lcf/rpg/learning.h
 	src/generated/lcf/rpg/map.h
 	src/generated/lcf/rpg/mapinfo.h
@@ -200,15 +185,15 @@ set(LCF_RPG_HEADERS
 	src/generated/lcf/rpg/music.h
 	src/generated/lcf/rpg/parameters.h
 	src/generated/lcf/rpg/rect.h
-	src/generated/lcf/rpg/save.h
 	src/generated/lcf/rpg/saveactor.h
 	src/generated/lcf/rpg/savecommonevent.h
 	src/generated/lcf/rpg/saveeasyrpgdata.h
 	src/generated/lcf/rpg/saveeventexecframe.h
 	src/generated/lcf/rpg/saveeventexecstate.h
+	src/generated/lcf/rpg/save.h
 	src/generated/lcf/rpg/saveinventory.h
-	src/generated/lcf/rpg/savemapevent.h
 	src/generated/lcf/rpg/savemapeventbase.h
+	src/generated/lcf/rpg/savemapevent.h
 	src/generated/lcf/rpg/savemapinfo.h
 	src/generated/lcf/rpg/savepanorama.h
 	src/generated/lcf/rpg/savepartylocation.h
@@ -230,27 +215,15 @@ set(LCF_RPG_HEADERS
 	src/generated/lcf/rpg/treemap.h
 	src/generated/lcf/rpg/troop.h
 	src/generated/lcf/rpg/troopmember.h
-	src/generated/lcf/rpg/trooppage.h
 	src/generated/lcf/rpg/trooppagecondition.h
+	src/generated/lcf/rpg/trooppage.h
 	src/generated/lcf/rpg/variable.h
 )
 
 list(TRANSFORM LCF_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
 list(APPEND LCF_HEADERS ${CMAKE_CURRENT_BINARY_DIR}/src/lcf/config.h)
 
-list(TRANSFORM LCF_LDB_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(TRANSFORM LCF_LMT_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(TRANSFORM LCF_LMU_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(TRANSFORM LCF_LSD_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(TRANSFORM LCF_RPG_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-
-target_sources(lcf PUBLIC ${LCF_HEADERS}
-	${LCF_LDB_HEADERS}
-	${LCF_LMT_HEADERS}
-	${LCF_LMU_HEADERS}
-	${LCF_LSD_HEADERS}
-	${LCF_RPG_HEADERS}
-)
+target_sources(lcf PUBLIC ${LCF_HEADERS})
 
 # IDE source grouping
 foreach(SG LDB LMT LMU LSD RPG)
@@ -349,12 +322,13 @@ install(
 	LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
 	RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
+foreach(S ldb lmt lmu lsd rpg)
+	set(SUBLIST ${LCF_HEADERS})
+	list(FILTER SUBLIST INCLUDE REGEX "/${S}/")
+	install(FILES ${SUBLIST} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/${S})
+	list(REMOVE_ITEM LCF_HEADERS ${SUBLIST})
+endforeach()
 install(FILES ${LCF_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf)
-install(FILES ${LCF_LDB_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/ldb)
-install(FILES ${LCF_LMT_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/lmt)
-install(FILES ${LCF_LMU_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/lmu)
-install(FILES ${LCF_LSD_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/lsd)
-install(FILES ${LCF_RPG_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/lcf/rpg)
 
 install(
 	FILES ${CMAKE_CURRENT_BINARY_DIR}/builds/${PROJECT_NAME}.pc

**Comment by fmatthew5876** [comment]
As long as we still explicitly define the full list of header files and don't use wild cards. I'm fine with regex / wildcard tricks after that to handle the installation paths.

**Comment by carstene1ns** [comment]
I think @Ghabry's approach looks a bit cleaner.
Source grouping is currently broken (needs regex update)
My script `sources2buildsystem.pl` should work with a small tweak

File: configure.ac [hunk 1]
```diff
@@ -11,7 +11,9 @@ AM_SILENT_RULES([yes])
 
 AC_CONFIG_MACRO_DIR([builds/autoconf/m4])
 AC_CONFIG_SRCDIR([src/data.cpp])
-AC_CONFIG_HEADERS([config.h])
+# We want our public lcf/config.h to look like the cmake one. This tricks autoheader into generating
+# /config.h (which we don't use) and then configure uses our template for lcf/config.h
+AC_CONFIG_HEADERS([autogen/config.h:config.h.in autogen/lcf/config.h:builds/autoconf.config.h.in])
```
carstene1ns, line 16: This should not really be needed, if we `AC_SUBST` the required variables and add the config file to `AC_CONFIG_FILES`
Then this line can be removed to not use config.h at all.

fmatthew5876: Thank you! This was ugly and I couldn't figure out the best way to handle it.

I will also change it to always define the macros to 0 or 1. The old style `#ifdef` way of detecting libraries is broken because you can forget to include the header.

Makefile.am
```diff
@@ -120,103 +135,106 @@ liblcf_la_SOURCES = \
...
-	src/generated/rpg_testbattler.h \
-	src/generated/rpg_treemap.h \
-	src/generated/rpg_troop.h \
-	src/generated/rpg_troopmember.h \
-	src/generated/rpg_trooppage.h \
-	src/generated/rpg_trooppagecondition.h \
-	src/generated/rpg_variable.h
+
+lcfinclude_HEADERS = \
+	autogen/lcf/config.h \
```
carstene1ns, line 140: This one should not be distributed, as is generated.

fmatthew5876: We need it to be installed when you do `make install` but we don't want it in a distribution tarball. Is there a way to get the former but not the latter?

carstene1ns: untested, but should be
```
nodist_lcfinclude_HEADERS = autogen/lcf/config.h
lcfinclude_HEADERS = \
```

carstene1ns:
```patch
diff --git a/Makefile.am b/Makefile.am
index 3479afa..035a003 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,7 +25,7 @@ lcfrpgincludedir = $(includedir)/lcf/rpg
 lib_LTLIBRARIES = liblcf.la
 liblcf_la_CPPFLAGS = \
 	$(AM_CPPFLAGS) \
-	-I$(srcdir)/autogen \
+	-I$(builddir)/autogen \
 	-I$(srcdir)/src \
 	-I$(srcdir)/src/generated
 liblcf_la_CXXFLAGS = \
@@ -136,8 +136,9 @@ liblcf_la_SOURCES = \
 	src/generated/rpg_state.cpp \
 	src/generated/rpg_system.cpp
 
+nodist_lcfinclude_HEADERS = \
+	autogen/lcf/config.h
 lcfinclude_HEADERS = \
-	autogen/lcf/config.h \
 	src/lcf/data.h \
 	src/lcf/encoder.h \
 	src/lcf/enum_tags.h \
```

fmatthew5876: Fixed, thanks

Ghabry:
liblcf-config.cmake is broken. Needs a different file in NAMES. e.g. "lcf/reader_lcf.h" works

PR #364, opened 2020-04-10
Add math optimization flags

fmatthew5876:


carstene1ns:
Do we even have enough math operations here that this matters?

Ghabry:
Right. There is only one cmath include in lsd_reader.cpp and it looks unused. (leftover from old timestamp code maybe)

fmatthew5876:
Practically we don't need it here. It just added it to keep in line with Player or in case we do add some floating point math later.

Up to you guys if we should merge or drop.

PR #366, opened 2020-04-23
Enable the GNU extensions

BSzili:
I'd like to get this nastiness:
https://github.com/EasyRPG/liblcf/blob/master/src/writer_xml.cpp#L12
The problem is what `-std=c++11` defines `__STRICT_ANSI__` which causes a lot of functions to go undeclared on AmigaOS4. I made this a separate PR in case it causes any problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ endif()
 # C++14 is required
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_EXTENSIONS ON)
 
 # lcf library files
 add_library(lcf
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,7 +21,7 @@ liblcf_la_CPPFLAGS = \
 	-I$(srcdir)/src \
 	-I$(srcdir)/src/generated
 liblcf_la_CXXFLAGS = \
-	-std=c++14 \
+	-std=gnu++14 \
 	-fno-math-errno \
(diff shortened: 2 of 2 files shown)


Ghabry:
Is it possible to update newlib?
They fixed these macros 4 years ago: https://github.com/bminor/newlib/commit/feec81e5717f904b687f17ace727bfbadc597209

BSzili:
I'm not part of the OS4 developer team, so I can't update it. I want to avoid hacking the SDK headers, because it would create a big mess, especially for others who would like to build liblcf, I might break other software, etc.
Unfortunately this seems to be the only way to be compatible with the official OS4 SDK without resorting to workarounds I added previously.

fdelapena:
In case we accidentally use some GNU extension in code, clang automated builds not based on autotools/cmake might fail, so maybe it's safe.

fmatthew5876:
I believe the only downside is that we if we use some gnu extension by mistake, non-gnu builds will fail. But then we'll see it in jenkins.

So I'm fine with this.

CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ endif()
 # C++14 is required
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_EXTENSIONS ON)
```
carstene1ns, line 30: ON is the default (could be removed).

fmatthew5876: I think it's better to be explicit in this case so we know what's going on.

Ghabry: I agree that being explicit here is better. Also makes git blame easier when the line is still there.

PR #367, opened 2020-04-26
Version 0.6.2

carstene1ns:
slow


builds/release-helper.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# release-helper.sh - maintainer utility script to change
+# the library version and update all copyright dates
+# by carstene1ns 2020, released under the MIT license
```
fmatthew5876, line 5: Doesn't look like the script will update itself for the year? :)

PR #369, opened 2020-05-15
Migrate lcf2xml to liblcf

Ghabry:
Don't review lcf2xml.cpp. The file is copied as-is. If you dislike it, submit patches later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ option(BUILD_SHARED_LIBS "Build shared library, disable for building the static 
 option(DISABLE_ICU "Disable ICU encoding detection (fallback to iconv, not recommended)" OFF)
 option(DISABLE_XML "Disable XML reading support (expat)" OFF)
 option(DISABLE_UPDATE_MIMEDB "Do not run update-mime-database after install" OFF)
+option(DISABLE_TOOLS "Do not build and install the tools" OFF)
 
 set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Override CMAKE_DEBUG_POSTFIX.")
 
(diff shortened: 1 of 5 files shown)


Ghabry:
I forward them now via AM_CONDITIONAL to Makefile.am but the checks are somehow broken.

I give up here. Somebody who understands this stuff can fix it. Sorry.

Makefile.am
```diff
@@ -240,6 +259,21 @@ test_runner_LDFLAGS = -no-install
 check-local:
 	$(AM_V_at)./test_runner
 
+lcf2xmldir = $(srcdir)/tools
```
carstene1ns, line 262: This line should not be needed.

Ghabry: Initially I needed this but now it works without ¯\\_(ツ)\_/¯

Ghabry:
ready. I will remove lcf2xml from tools (and fix the rest)

CMakeLists.txt
```diff
@@ -374,3 +375,19 @@ foreach(i ${BENCH_FILES})
 	add_executable(bench_${name} ${i})
 	target_link_libraries(bench_${name} lcf)
 endforeach()
+
+# tools
+if(NOT DISABLE_TOOLS)
+	set(TOOLS lcf2xml)
```
carstene1ns, line 381: I think it would be cleaner to do it the other way around:
Empty tool list, add lcf2xml when not disabled.

Makefile.am
```diff
@@ -259,6 +259,20 @@ test_runner_LDFLAGS = -no-install
 check-local:
 	$(AM_V_at)./test_runner
 
+lcf2xml_SOURCES = tools/lcf2xml.cpp
+lcf2xml_CPPFLAGS = $(liblcf_la_CPPFLAGS)
+lcf2xml_CXXFLAGS = $(liblcf_la_CXXFLAGS)
+lcf2xml_LDADD = liblcf.la $(liblcf_la_LIBADD)
```
carstene1ns, line 265: This looks like it overlinks with unneeded icu.
Also should be enough to link with lcf really, libtool should figure out the rest.

(needs testing)

Ghabry: works without

Issue #368, opened 2020-05-14
Make all RPG types printable


fmatthew5876:
Implement `operator<<` for all the rpg types so we can print them using fmtlib logging.

Starting this is blocked by https://github.com/EasyRPG/liblcf/pull/361

PR #372, opened 2020-05-26
Make rpg structs printable

fmatthew5876:
The objective is add a simple mechanism to print out lcf structs to aid debugging.

- [x] Test with player fmtlib output logger

Fix: #368 

Examples:
```
SavePartyLocation{boarding=0, aboard=0, vehicle=0, unboarding=0, preboard_move_speed=4, menu_calling=0, pan_state=1, pan_current_x=2304, pan_current_y=1792, pan_finish_x=2304, pan_finish_y=1792, pan_speed=16, encounter_steps=0, encounter_calling=0, map_save_count=0, database_save_count=0}
```

```
/mnt/c/Users/matt/Documents/dev/Player/tests/move_route.cpp:635:
TEST SUITE: MoveRoute
TEST CASE:  CommandChangeGraphic

/mnt/c/Users/matt/Documents/dev/Player/tests/move_route.cpp:99: FATAL ERROR: REQUIRE_EQ( ch.IsMoveRouteRepeated(), done ) is NOT correct!
  values: REQUIRE_EQ( false, true )
  logged: paused := false
          move_frequency := 2
          stop_count := 65536
          max_stop_count := 128
          move_route_idx := 1
          overwritten := false
          done := true
          mr := MoveRoute{move_commands=[MoveCommand{command_id=34, parameter_string=x, parameter_a=3, parameter_b=0, parameter_c=0}], repeat=0, skippable=0}
```


fmatthew5876:
There is an open question as to where the implementations should live.

We don't want these inlined in the headers. That will cost in compile time. I opted to put these in rpg_NAME.cpp. The issue though is before we only had a few of these rpg source files, now we have 1 for every struct.

fmatthew5876:
jenkins: test this please

fmatthew5876:
Any feedback on this? I'd like to get this in so my printing of lcf structs compiles in player prs. If there is big disagreement on the output format maybe we can punt that? I'm happy to change this later, just want `operator<<()` to compile for now.

Ghabry:
As long as this is really only for debugging printing and I don't see stuff like string comparing of lcf types in the future I don't really mind.
It will make building liblcf slightly slower but at least won't affect Player.

fmatthew5876:
Yes only for debuging, and no this should not affect player build times. Thats the reason for all the new cpp files.

CMakeLists.txt
```diff
@@ -124,11 +124,72 @@ target_sources(lcf
...
+	src/generated/rpg_animationframe.cpp
+	src/generated/rpg_animationtiming.cpp
+	src/generated/rpg_attribute.cpp
+	src/generated/rpg_battlecommand.cpp
+	src/generated/rpg_battlecommands.cpp
+	src/generated/rpg_battleranimation.cpp
+	src/generated/rpg_battleranimationdata.cpp
+	src/generated/rpg_battleranimationextension.cpp
 	src/generated/rpg_chipset.cpp
-	src/generated/rpg_enums.cpp
```
carstene1ns, line 128: gone?

fmatthew5876: Really good catch. This would only manifest as a linker error for debug builds on some platforms. Fixed.

Issue #169, opened 2016-02-23
Do lazy reading instead of reading the whole file twice on init


fdelapena:
Some games have large database files and it can be slow parsing it completely.

Furthermore, the file is parsed twice: first to get some strings to detect the encoding with ICU, and then parsed again with the detected encoding.

For a proof of concept, try the Hero's Realm HTML5 version:
https://easy-rpg.org/play/index-master.html?game=heros_realm

It should allow to do selective parsing of data instead of eager loading.

Ghabry:
Could be combined with a reader_struct refactor ;)

fmatthew5876:
Lazy reading incurs a branch on every read because you always have to check if it's loaded or not. Most things in the database are used frequently enough you'd want them in main memory all the time.

One way we could speed up load times is by tricking the user. When starting up Player, we could load only the bare minimum required the operate the title screen and load menu. Then, while the user is on the title screen, we continue loading the rest of the database (and maybe the maps of his saves in order of most recent save) in a background thread. By the time the user picks new game or loads, a lot of, if not all of the lcf reading is done.

fdelapena:
Since 0.6.2, games load even faster than RPG_RT. Maybe this issue can be closed.

fmatthew5876:
Reading everything twice is still pretty bad and fixing this could speed up significantly more.

Maybe it's possible to do something which only loads enough partial string data to figure out encoding and exits.

I'd like to leave this open until address the double load issue.

Ghabry:
this will be solved when Saves use DBString, before we can't get rid of the Encode functions in the parsing code.

PR #388, opened 2020-08-25
Add ForEachString

Ghabry:
This adds a new function ``ForEachString`` for replacing strings in LCF. Used for translation as discussed in that Player PR.

CC @sorlok

### DIFF
--- a/generator/generate.py
+++ b/generator/generate.py
@@ -150,6 +150,11 @@ def filter_size_fields(fields):
     for field in fields:
         if not field.size:
             yield field
+
+def filter_field_has_type(fields, typ):
+    for field in fields:
+        if field.type == typ:
+            yield field
 # End of Jinja 2 functions
 
 int_types = {
@@ -463,6 +468,7 @@ def main(argv):
     env.filters["field_is_used"] = filter_unused_fields
     env.filters["field_is_written"] = filter_unwritten_fields
     env.filters["field_is_not_size"] = filter_size_fields
+    env.filters["field_has_type"] = filter_field_has_type
     env.filters["num_flags"] = num_flags
     env.filters["flag_size"] = flag_size
     env.filters["flag_set"] = flag_set
--- a/generator/templates/rpg_header.tmpl
+++ b/generator/templates/rpg_header.tmpl
@@ -87,6 +87,9 @@ namespace rpg {
 		{%- endif %}
 	{%- endif %}
 	{%- endfor %}
+
+		template <typename F>
+		void ForEachString(const F& f);
 	};
 	{%- if struct_name in enums %}
 	{%- for name, enum in enums[struct_name].items() %}
@@ -128,6 +131,15 @@ namespace rpg {
 	}
 
 	std::ostream& operator<<(std::ostream& os, const {{ struct_name }}& obj);
+
+	template <typename F>
+	inline void {{ struct_name }}::ForEachString(const F& f) {
+	{%- for field in fields[struct_name]|field_is_used|field_has_type("String") %}
+		{{ field.name }} = f({{ field.name }}, "{{ field.name }}");
+	{%- else %}
+		(void)f;
+	{%- endfor %}
+	}
 } // namespace rpg
 } // namespace lcf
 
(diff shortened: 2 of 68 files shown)


fmatthew5876:
1. I would change the lambda to not have a return value, then I can write these:


```
[](auto& field, const auto& name) { field = transform(field); };
[](auto& field, const auto& name) { Utils::ToUpperInplace(field); };
[](const auto& field, const auto& name) { doSomethingReadOnly(field); };
```

This also grants some potential efficiency as you can modify the already allocated string object in place and not create a new one with a return value.


2. Also, the name is not enough information. Inside my lambda, you give me the name of the field but I don't which struct it is. I would also pass in the struct itself.

```
f(*this, name, "name");
```

3. Consider making `ForEachString()` a free function overloaded for all rpg types instead of a member function. That makes it easier to extend to other types.

Ghabry:
What is "a free function overloaded for all rpg types". (I only know overloading in the context of class inheritance)

Do you mean a template function with specialization per rpg type?

fmatthew5876:
```
template <typename F>
void ForEachString(lcf::rpg::SaveActor&, const F& f);
template <typename F>
void ForEachString(lcf::rpg::SaveMapEvent&, const F& f);
```
 etc..

Don't template the struct type, C++ already gives us function overloading to do that.

Ghabry:
Another use case: Reencoding the entire database without parsing again (at least once, for "Unknown -> Detected")

sorlok:
So right now you get the property's name ("end_game") and the value ("Shutdown"). When looking this up in the .po files, we actually look up by context ("term"). See sample code below.

```
	// Translate all terms
	lcf::Data::terms.ForEachString([this](const std::string& value, const std::string& name)->std::string {
		if (sys!=nullptr) {
			return sys->translate_ctxt("term", value);
		}
		return value;
	});
```

I think the following changes make sense:
1) Pass the context ("term") in to the callback function as well. This streamlines translation substitution.
2) Switch the order of name and value (stylistic)
3) Pass the value in by NON-const reference, so we don't *have* to always overwrite it.

fmatthew5876:
> Another use case: Reencoding the entire database without parsing again (at least once, for "Unknown -> Detected")

So you mean we don't have to load the entire database twice?? That'll be a huge win for player!

sorlok:
One other thing to consider (in addition to my comment above)  ---right now, if you switch from Default language to Spanish, then from Spanish to French, I have to reload the database (so that we can look up strings properly). I don't think this is worth fixing, since switching between multiple languages at runtime is an edge use case. Just to make sure everyone's aware though.

Ghabry:
> So you mean we don't have to load the entire database twice?? That'll be a huge startup time performance win for player!

Yeah exactly this. See #169

> So right now you get the property's name ("end_game") and the value ("Shutdown"). When looking this up in the .po files, we actually look up by context ("term"). See sample code below.

matthew already suggested to also provide an object instance. You can use the object instance to get a name through "reflection": ``Struct<std::remove_reference_t<decltype(instance)>>::name``. But this is 1. not public API and 2. uglyyyy. Maybe add some ``lcf/reflection.h`` header which abstracts this stuff (``GetTypename(rpg::instance&)``, also add ``ForEachString`` there), then we can freely alter the templates later without breaking code.
CC @fmatthew5876 You are the template master here (I needed 10 minutes to make this ``Struct`` line compile ;)). thoughts?

sorlok:
The "reflection" idea sounds fine to me. I also like that LcfTrans can be updated to always be in sync on these contexts.

Is there any case where the name of the "context" (passed in to getText()) would need to be very different from the name of the class? That's the only thing I can think of that might trip this up. So far, I can't think of any case where this would be true.

fmatthew5876:
`Struct<std::remove_reference_t<decltype(instance)>>::name`

Reader struct is not part of the public API, and it never should be. 

What I think we need here is a new reflection mechanism that is public, and refactor reader_struct to use it.

sorlok:
Hi, just checking in. Should I wait for the refactoring to ForEachString() or continue coding with the current API? (This is for the Translation PR I'm working on.) I've still got lots of side stuff to do, so no rush.

fmatthew5876:
I looked into proper reflection, it got way over complicated too fast. Also compile times doubled.

Since our main problem is just being able to translate strings, I suggest we move forward with this for now. You could just pass in the struct name and member name as string literals.

Ghabry:
closing in favor of #394

Ghabry:
@sorlok 

Please wait a bit more. I will ping you when it is ready.

PR #394, opened 2020-09-09
Add ForEachString

fmatthew5876:
Here is a prototype for a recursive `ForEachString` function. Also included is a simple tool to dump all strings in an lcf file, which shows how to use it.

I tried to do a bunch of fancy reflection, but it got way too over complicated and make compile times nearly double. Since all we really care about right now is being able to do translations, lets not over-engineer.

@sorlok How are you planning to identify strings and map them between languages? What file format will you use to store translations?

Right now this just passes struct name and field name which is obviously insufficient.

How will you identify something like "the string of the 5th event command in the 2nd page of event 3 on map 5" etc..


fmatthew5876:
I just pushed a change which gets rid of the struct / field names and uses these as a context identifier:

```
::troops[8]::pages[11]::event_commands[263]::string = "Trigger STEAL event to take the item."
::troops[8]::pages[12]::event_commands[0]::string = "VERSION 1 TROOPSIZE 3"
::troops[8]::pages[12]::event_commands[1]::string = "vvvvvvvvvvvvvvvvvvvvvvvvvvvvvv"
::troops[8]::pages[12]::event_commands[2]::string = "vvvvvvvvvvvvvvvvvvvvvvvvvvvvvv"
::troops[8]::pages[12]::event_commands[4]::string = "Which character?"
::troops[8]::pages[12]::event_commands[6]::string = "Battle Char 1 - who is trying to sneak"
::troops[8]::pages[12]::event_commands[7]::string = "Battle Char 2 - Address of SneakRounds counter"
::troops[8]::pages[12]::event_commands[8]::string = "Battle Local 1 - char AGI"
::troops[8]::pages[12]::event_commands[9]::string = "Battle Local 3 - sneak bonus"
::troops[8]::pages[12]::event_commands[10]::string = "Battle Local 14 - SneakRounds value"
::troops[8]::pages[12]::event_commands[33]::string = "\>Char \n[\v[996]] Rounds Var[\v[997]] = \v[994]"
::troops[8]::pages[12]::event_commands[34]::string = "AGI = \v[981]"
::troops[8]::pages[12]::event_commands[35]::string = "Bonus Mod = \v[983]"
::troops[8]::pages[12]::event_commands[36]::string = "\<"                                                                                                                                                                                    ::troops[8]::pages[12]::event_commands[39]::string = "Accumulate bonus from Map Visbility level"
::troops[8]::pages[12]::event_commands[41]::string = "Menu Local 2 - Cum Bonus"
::troops[8]::pages[12]::event_commands[45]::string = "\>Map Mod = \v[3]"                                                                                                                                                                     ::troops[8]::pages[12]::event_commands[46]::string = "Cum Mod = \v[983]\<"
::troops[8]::pages[12]::event_commands[49]::string = "Get max enemy AGI"
::troops[8]::pages[12]::event_commands[50]::string = "<ENEMY SPECIFIC CODE HERE>"
::troops[8]::pages[12]::event_commands[52]::string = "Battle Local 6 - Which enemy"                             
```

Ghabry:
Here is the current output of the old lcftrans: https://gist.github.com/Ghabry/7c23350f5787bd1af20fd34602919a95

It creates 3 files for the DB:
- RPG_RT.po: The database without events
- RPG_RT_battle.po: The battle events
- RPG_RT_common.po: The common events

And N fils for the map:
- MapXXXX.po: All strings on the map

About the structure:
In a po file the ``msgid`` is always the source language string. ``msgstr`` must contain the translated string. (There are also plural forms but the plural rule depends on the target language and is out-of-scope for now)

``msgctx``: This is for handling homonyms. E.g. right can mean the direction or correct, in German this would be "rechts" and "richtig"

There is currently _no way_ to update an existing translation e.g. when a game was updated and the database changed. This is a hard problem and out-of-scope for the first version imo.

---------

In your tool you have now (excerpt from VD1):

```
::actors[0]::name = "Valnar"
::actors[0]::title = "Krieger"
::actors[1]::name = "Asgar"
::actors[1]::title = "Vampir"
```

The translators are not programmers: _Counting should start from 1_

The reason why I didn't want theo entire context construction in liblcf is flexibility. E.g. ForEachString is invoked and passes ("actors", 1, "name"), then lcftrans will lookup first for context ``actors.name.1`` and falls back to ``actors.name``. When you pass ``::actors[0]::name`` there is lots of string parsing involved.

E.g. lcftrans emitted this (#. is a comment)

```
#. Actor 1: Name
#. Actor 11: Name
msgctxt "actors.name"
msgid "Valnar"
msgstr ""
```

So actor 1 and 11 have the same name: You only have to translate it once...

-----------------

For events the callback-approach will not work. This must be done during messagebox construction.
lcftrans parses the messages and combines them to one string.

Common Events. The following lines appears lots of times so it should only be translated once:

```
#. Common Event 1, Line 6
#. Common Event 1, Line 15
#. Common Event 1, Line 24
[...]
#. Common Event 9, Line 15
#. Common Event 9, Line 24
#. Common Event 9, Line 33
msgid ""
"\\c[3]Asgar:\\c[0]\n"
"Beschwörungen funktionieren nur\n"
"in unverwandelter Gestalt !"
msgstr ""
```

There is no context but for a more specific match the lookup could be ``1.6`` (CE1, Line 6) -> ``1`` (CE1) -> ```` (Global)

As you can see there is no ``msgctx`` at all. The idea is: When there are homonyms the translator has to manually duplicate the entry and provide a context.

-----------------

Maps:

E.g. Map0001 has 4 times the same message box.

```
#. Event 9, Page 1, Line 3
#. Event 12, Page 1, Line 3
#. Event 13, Page 1, Line 3
#. Event 14, Page 1, Line 3
msgid ""
"\\c[3]Asgar:\\c[0]\n"
"Wir sollten zuerst zu der Mine gehen !"
msgstr ""
```

As you can see there is no ``msgctx`` at all. The idea is: When there are homonyms the translator has to manually duplicate the entry and provide a context.
No idea how the context could look like, in your code
```
::events[8]::pages[0]::event_commands[2]::string = "\c[3]Asgar:\c[0]"
```

The lookup order could be: ``9.1.3`` (Event+Page+Line) -> ``9.1`` (Event+Page) -> ``9`` (Event) -> ```` (Global)

I consider it relevant that a translator can manually type the context.

-----------

Other stuff:
- #: Source line reference. Not supported yet / not useful
- #, Fuzzy indicator (when a Po file was updated based on an updated sourcecode file. Not supported yet)

Ghabry:
With some patching I can get which looks better imo. (this is like Django template syntax)

```
actors.1.name = "Valnar"
actors.1.title = "Krieger"
actors.1.character_name = "UnChar4"
actors.1.face_name = "Vampire Faces"
actors.2.name = "Asgar"
actors.2.title = "Vampir"
actors.2.character_name = "Extra2"
actors.2.face_name = "Monsters"
[...]
skills.11.description = "Sehr starker Zerstörungszauber trifft einen Gegner"
skills.11.using_message1 = " zaubert Apokalypse !"
skills.11.sound_effect.name = "(OFF)"
[...]
terms.encounter = " erscheint !"
terms.special_combat = "Ihr überrascht den Gegner !"
terms.escape_success = "Ihr flieht in Schande ..."
[...]
events.9.name = "EV0009"
events.9.pages.1.event_commands.2.string = "Vampire Faces"
events.9.pages.1.event_commands.3.string = "\c[3]Asgar:\c[0]"
events.9.pages.1.event_commands.4.string = "Wir sollten zuerst zu der Mine gehen !"
```
(Though for events there will be manual processing needed anyway)

```
diff --git a/generator/templates/rpg_header.tmpl b/generator/templates/rpg_header.tmpl
index 4206029..c5e6c2b 100644
--- a/generator/templates/rpg_header.tmpl
+++ b/generator/templates/rpg_header.tmpl
@@ -131,14 +131,15 @@ namespace rpg {
 
 	template <typename F>
 	void ForEachString({{ struct_name }}& obj, const std::string& ctx, const F& f) {
+		std::string p = ctx.empty() ? "" : ".";
 		{%- for field in fields[struct_name]|field_is_used|field_is_not_size -%}
 		{%- if field.type is is_string %}
-		f(obj, ctx + "::{{ field.name }}", obj.{{ field.name }});
+		f(obj, ctx + p + "{{ field.name }}", obj.{{ field.name }});
 		{%- elif field.type is is_struct %}
-		ForEachString(obj.{{field.name}}, ctx + "::{{ field.name }}", f);
+		ForEachString(obj.{{field.name}}, ctx + p + "{{ field.name }}", f);
 		{%- elif field.type is is_array_of_struct %}
 		for (int i = 0; i < static_cast<int>(obj.{{ field.name }}.size()); ++i) {
-			ForEachString(obj.{{ field.name }}[i], ctx + "::{{ field.name }}[" + std::to_string(i) + "]", f);
+			ForEachString(obj.{{ field.name }}[i], ctx + p + "{{ field.name }}." + std::to_string(i+1), f);
 		}
 		{%- endif -%}
 		{%- endfor %}
```

fmatthew5876:
Feel free to take this over and change it however makes the most sense for lcftrans

sorlok:
Thanks for the prototype @fmatthew5876 (and the detailed response @Ghabry ).

From the point of view of translations, I think we only need *basic* contexts, and we can do string matching based on the source language string (i.e., just basic gettext stuff). I would recommend passing in just "actors.name" or "::actors::name"  instead of "::actors[0]::name".

However, it seems like "ForEachString()" might have other uses outside translation, and for those uses, the exact actor ID (message ID, etc) might be useful. Is that a correct assessment, or do we think actor IDs will be useful for translations? 

Assuming we have multiple uses of ForEachString(), I wonder if we can abstract away what isn't needed for translations. In other words, I'd rather not have to parse "::actors[0]::name" down into "::actors::name". I hate to suggest it, but maybe we need two "context" strings: detailed and simple (and the "detailed" one can be empty for now)? Or maybe just go with the simple context to start and fill in new information when we actually have an application that uses ForEachString() and needs it?

Ghabry:
@sorlok 

I can see currently three use cases to cover:
- Generating a translation file, translating strings (that's almost 2 cases)
- Reencoding strings (no context needed here)
- Renaming files (not for translating! For tools such as "Change as change Japanese files to ASCII")

For case 1.1 and 3 the event ID is highly useful, fortunately it is already in the object instance.

As said I think for a lookup using two contexts is better: First a specific lookup (e.g. actors.1.name), then a generic lookup (e.g. actors.name). For covering homonym/autoantonym cases I can't currently think off.

I already think about passing in some "Context info" struct with the following data:
- lcf::Context* prev <- The context of the parent
- void* obj <- well... cast based on obj name
- StringView obj_name
- StringView field_name
- int index
- StringView field_value

Ghabry:
Well I overengineered this now a bit but is more flexible:

```
./lcfstrings RPG_RT.ldb 1252
actors[0]::name=Valnar
actors[0]::title=Vampir
actors[0]::character_name=VDChar
actors[0]::face_name=1-valnar
actors[0]::skill_name=Sorcery
actors[1]::name=Asgar
actors[1]::title=Vampir

./lcfstrings Map0001.lmu 1252
events[48]::name=EV0049
events[49]::name=EV0050
events[50]::name=EV0051
events[50]::pages[0]::event_commands[0]::string=VDFaces2
events[50]::pages[0]::event_commands[2]::string=\c[10]Sir Aaron:\c[0]
events[50]::pages[0]::event_commands[3]::string=Burg Horan liegt südlich von hier!

./lcfstrings -1 RPG_RT.ldb 1252
actors[1]::name=Valnar
actors[1]::title=Vampir
actors[1]::character_name=VDChar
actors[1]::face_name=1-valnar
actors[1]::skill_name=Sorcery
actors[2]::name=Asgar
actors[2]::title=Vampir

./lcfstrings -1 Map0001.lmu 1252
events[49]::name=EV0049
events[50]::name=EV0050
events[51]::name=EV0051
events[51]::pages[1]::event_commands[1]::string=VDFaces2
events[51]::pages[1]::event_commands[3]::string=\c[10]Sir Aaron:\c[0]
events[51]::pages[1]::event_commands[4]::string=Burg Horan liegt südlich von hier!

./lcfstrings -s Map0001.lmu 1252
events.48.name=EV0048
events.49.name=EV0049
events.50.name=EV0050
events.51.name=EV0051
events.51.pages.1.event_commands.1.string=VDFaces2
events.51.pages.1.event_commands.3.string=\c[10]Sir Aaron:\c[0]
events.51.pages.1.event_commands.4.string=Burg Horan liegt südlich von hier!

./lcfstrings -s Map0001.lmu 1252
actors.0.name=Valnar
actors.0.title=Vampir
actors.0.character_name=VDChar
actors.0.face_name=1-valnar
actors.0.skill_name=Sorcery
actors.1.name=Asgar
actors.1.title=Vampir

./lcfstrings -s -1 Map0001.lmu 1252
actors.1.name=Valnar
actors.1.title=Vampir
actors.1.character_name=VDChar
actors.1.face_name=1-valnar
actors.1.skill_name=Sorcery
actors.2.name=Asgar
actors.2.title=Vampir
```

fmatthew5876:
I think I made this mistake too when considering general reflection. For now lets stay focused on the translation use case and deliver that. If we need a more general facility later, we can always refactor it.

Adding more generalization / complexity later is easy, taking away over engineered complexity later is hard.

Ghabry:
I exaggerated a bit with "overengineered". I mean this lcf::Context idea is now more complex than your approach.

For simplicity reasons it only works for DBString right now (unusable for savegames) but this will be solved by other work in the next weeks and translating of saves is not relevant.

But it solves:

----------

- Translating strings

Context making:

```cpp
if (ctx.parent->index == -1) {
    ctx = ctx.parent->name + "." + ctx.name
} else {
    ctx_specific = ctx.parent->name + "." + (ctx.parent->index+1) + "." + ctx.name
    ctx_generic = ctx.parent->name + "." + ctx.name
}

```

------------

- Generating a translation file (lcftrans)

Checking fo "ctx.parent->name == eventcommands" and casting the ``void*`` to rpg::eventcommand and fetching the event code. Can be used for msg command processing in lcftrans. Not a big fan of this void-casting but otherwise you need a template arg for the parent, quite annoying because you usually won't access it.

-----------

- Reencoding strings (no context needed here)

Possible (no tool available)

out-of-scope: Move command filenames

-----------

- Renaming files (not for translating! For tools such as "Change as change Japanese files to ASCII")

Possible through context (no tool available)

out-of-scope: Move command filenames

----------

Out-of-scope ideas:

By adding more ``ForEachObject`` methods the entire database could be processed via callbacks (visitor pattern).
XML, JSON, etc. transformations could be made through this instead of hardcoding it in liblcf.

sorlok:
Thanks for the refactoring. I'll write some code tonight/tomorrow to test it out.

Is it correct to say that I want my translation-applying code to do the following? (For an actor's name)
1. Check for a translation with "actors.0.name" as the context
2. If that fails, check for a translation with "actors.name" as the context
3. If that fails, check for a translation with no context
4. If that fails, just use the original language string

Quick questions:
1. Will LCFTrans be updated to generate e.g., "events[51]::pages[1]::event_commands[3]::string" as a context for the message boxes within map events? Or will the context for message boxes still be empty?
2. How do I deal with "events[50]::pages[0]::event_commands[0]::string" for step 2 of the previous paragraph? I.e., will it just become "events[50]::pages[0]::event_commands::string", or am I expected to recurse all the way down to "events::pages::event_commands::string"?
3. Are the translation authors expected to manually change, e.g., "actors.0.name" to "actors.name" inside their .po files in cases where they want to share string (think: 10 characters named Alex) ---or is it the opposite (they will be given "actors.name" by default and will have to manually add "actors.0.name" if that Alex is different for some reason)?

Thanks!

Ghabry:
FYI to test this PR I made a quick-n-dirty updated lcftrans that ONLY extracts the Database strings in a po (events are unsupported). Maybe good for testing your code? :+1: 

https://github.com/Ghabry/easyrpg-tools/tree/lcftrans2

Usage of ForEach: https://github.com/Ghabry/easyrpg-tools/blob/lcftrans2/lcftrans/src/translation.cpp#L154
(Note that "*ctx.value" is writable, just assign the lcf::DBString to it)

Is quite nice. I can do now in ~40 lines (bit more when finished) what I did before in 800 lines.

---------------

> Is it correct to say that I want my translation-applying code to do the following? (For an actor's name)

> Check for a translation with "actors.0.name" as the context

Yes. Add 1 to the index because the translators are not programmers

> If that fails, check for a translation with "actors.name" as the context

Yes

> If that fails, check for a translation with no context

I didn't consider this but: Yes

> If that fails, just use the original language string

Yes

--------------

> Will LCFTrans be updated to generate e.g., "events[51]::pages[1]::event_commands[3]::string" as a context for the message boxes within map events? Or will the context for message boxes still be empty?

Havn't fully decided this yet. Likely will be context-free by default and for a better match they can add a context.
I don't think hotpatching event commands is feasible. Later we maybe need injecting more text lines, this is not possibel with this patch approach. Better "hook" into ShowMessage/ShowChoice.

> Are the translation authors expected to manually change, e.g., "actors.0.name" to "actors.name" inside their .po files in cases where they want to share string (think: 10 characters named Alex) ---or is it the opposite

It's the opposite. The more common case is the default one. For the same context (e.g. actors.skill) I expect that in 99% of the cases that same word = same translation.

sorlok:
Ok, sounds good. This approach actually seems very generic and easy to understand (but still possible to customize), so I like it. I'll post back when I finish updating my stuff (thanks for the sample lcftrans2).

sorlok:
The branch has been updated with code that uses the new ForEach for Terms (well, for "lcf::Data::data", but I've only tested terms). I'll be expanding it and cleaning it up over the next few days.

Two questions:
1. For non-message boxes, do I ever have to care about more than one parent per context? I.e., "actors.name" is fine, but will I ever get "party.actors.name" or something like that? (I know that for event + message boxes we are not using ForEach).
2. Do we want utility functions for computing the "full" and "concise" lookup strings? Otherwise, I might choose to write "actors.0.name" while you write "actors[0].name".

Ghabry:
I'm not sure if I missed something but for non-event-stuff everything appears to be one level deep.

> For non-message boxes, do I ever have to care about more than one parent per context

That's the reason for this check: 
```
if (!ctx.parent || ctx.parent->parent) {
	// Only care about entries one level deep
	return;
}
```

> Do we want utility functions for computing the "full" and "concise" lookup strings

I don't want to put the translation stuff in liblcf currently. I bet we will encounter problematic edge cases we didn't think of yet. (this includes the lcftrans tool). When the first games are translated and the biggest problems resolved we can still merge it later :) 

In Player code you can use fmtlib btw. ``fmt::format`` looks much better than having lots of ``+``: https://fmt.dev/latest/index.html

```cpp
const auto& p = ctx.parent;

// With index:
fmt::format("{}.{}.{}", p.name, p.index + 1, ctx.name)

// Without index:
fmt::format("{}.{}", p.name, ctx.name)
```

Ghabry (approved):
This is now flexible enough for use in LcfTrans and for translating of strings in Player so I give my approval here.

sorlok:
I'm happy with the API; it will work for my Translation stuff just fine.

src/lcf/context.h
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Context information for lcf fields
+ * @tparam T field type
+ */
+template <typename T>
+struct Context {
+	/** Context of the parent (nullptr when no parent) */
+	Context<T>* parent = nullptr;
+	/** Object instance (cast to appropriate RPG-type */
+	void* obj = nullptr;
```
fmatthew5876, line 26: We're already passing obj to the function directly, why do we need this void pointer?

Ghabry: This is used when you want to reference the instance of the parent object.
E.g. when I parse a map and the parser is on the string of an event command I can access the "lcf::rpg::Event" with 
``auto* event = reinterpret_cast<lcf::rpg::Event*>(ctx.parent->parent->obj)``

fmatthew5876: Ok, I think I can add a little more template fairy dust and get rid of the need for the reinterpret_cast here. Do we want that?

Ghabry: Well I couldn't figure out the template magic to get rid of the void* without making the API too ugly. But feel free to improve it :).

fmatthew5876:
So now we have 3 implementations. Each one giving more type safety at the cost of more template complexity.

1. Ghabry's implementation using void*. Kind of clunky and error prone requiring `reinterpret_cast`. Notice how in lcfstrings however the print functions don't have to be templated.
2. Add string type for the current context, and void for all parents. This gives us type safety at one level where we need it, all parents struct types get erased to `void*`
3. Fully recursive. Now we have `Context<DBString,Struct,Context<DBString,ParentStruct,Context<DBString,GrandParentStruct,void>>>` etc.. You get type safety all the way up. Now more templates everywhere and you have to handle the case where the top level parent is a void* and not a context. (see `print_parent()` in lcfstrings)

I'll let you guys decide which one you prefer :)

fmatthew5876:
I made another minor improvement.

First, we should not store the actual field in the context itself. It doesn't make sense when all the parent contexts have a null field. So I took it out, and removed one template parameter in the process.

Also I made the context passed in as const, since it makes no sense to try to modify it from your callback.

I actually quite like this. Later on, we could potentially generalize this as a new construct for iterating over all structs and fields. That machinery could become the basis of a revamp of the lcf and xml parsers, and also allow anyone to implement custom file formats by simply iterating fields.

Ghabry:
I like the typesafe
```
[](const auto& val,
	lcf::Context<lcf::rpg::EventCommand,
	lcf::Context<lcf::rpg::EventPage,
	lcf::Context<lcf::rpg::Event, void>>>& ctx)
```

Now that the value is an argument again the limitation on DBString can be lifted and std::string supported again. My only reason why I removed it is that it caused too much template chaos but you found a way to solve it :). (it really made no sense to make the instance part of the context)

Ghabry:
I think we need a lcf::rpg::IDontCare (Any?) type.

For LcfTrans when dumping strings from the database I only care about the parent name and index for filtering the strings I want to dump (so the object instance doesn't matter). I can't use "void" here but I also don't want to "lie" by using a random lcf type.

sorlok:
Cool iterations on the function! I'll update my code tomorrow and see how it goes.

fmatthew5876:
> Now that the value is an argument again the limitation on DBString can be lifted and std::string supported again. 

I'm tempted to leave it alone. Once the other refactor PRs get merged I plan to use DBString everywhere so then it's no longer relevant. It almost makes writing translation code easier if you only have to deal with 1 string type.

> I think we need a lcf::rpg::IDontCare (Any?) type. _EDIT: Well I can just declare an empty struct anywhere and use this as the type. So doesn't really matter_

I'm not sure how useful this really is, but I recently had an idea to have an inheritance hierarchy for all rpg types.

`struct BaseObject {};` which is empty and the base of everything
`struct Object : BaseObject { int ID; }` contains only ID and is the base of all rpg types with ids.

This isn't that useful because there would be no virtual functions, and we don't ever want to add vtables as that will cost a lot in memory.

> 
> For LcfTrans when dumping strings from the database I only care about the parent name and index for filtering the strings I want to dump (so the object instance doesn't matter). I can't use "void" here but I also don't want to "lie" by using a random lcf type.

We can support that with a base context class type which doesn't have any template parameters and no type information.

Basically if `Context<T,Parent>` inherits from `ContextBase` and ContextBase only has the strings and ints, you can get out of template land by upcasting to the base class.

Not sure how useful this is, but I also had an idea of how maybe you could "strip" the parent context, something like:

```
lcf::Context<lcf::rpg::EventCommand,lcf::Context<lcf::rpg::EventPage,lcf::Context<lcf::rpg::Event, void>>> x;
lcf::Context<lcf::rpg::EventCommand, void> y = stripParents(x);

**Comment by fmatthew5876** [comment]
Ok I added some inheritance to context so you can drop the type info. This can be used to improve compile and run time performance.

Take a look at `print_ctx_info` in `lcfstrings.cpp`. While there are many copies of `print_parent`, one per template instantiation, there is only 1 copy of `print_ctx_info` compiled.

This helps at runtime too, because one copy of the code means better icache behavior.

**Comment by Ghabry** [comment]
More funny problems with typing.
I want to process all (common) events and for the choice extraction I need the indent. I can't use Database::commonevents as a starting point because this is a vector so the starting point is the owner of the vector: The entire Database.
This fails to compile because rpg::Actor, Skills etc. has no attribute "indent".

```cpp
lcf::rpg::ForEachString(lcf::Data::data, [&](auto& val, auto& ctx) {
    /* Some lines omitted */
    // is not always rpg::EventCommand -> error
    int indent = ctx.obj->indent;
    /* */
}
```

What is the best way to handle this?
My current verbose solution is an overloaded function:

```cpp
template <typename T, typename U, typename V>
void processCommonEvent(const T&, const lcf::Context<U, V>&) {}

template <typename T, typename V>
void processCommonEvent(const T& val, const lcf::Context<lcf::rpg::EventCommand, V>& ctx) {
	// is rpg::EventCommand and compiles
        int indent = ctx.obj->indent;
}

lcf::rpg::ForEachString(lcf::Data::data, [&](auto& val, auto& ctx) {
    processCommonEvent(val, ctx);
}
```

fmatthew5876:
> More funny problems with typing.
> I want to process all (common) events and for the choice extraction I need the indent.
> This fails to compile because rpg::Actor, Skills etc. has no attribute "indent".
> What is the best way to handle this?


Basically the way you're doing it is the right way, however if you don't need parent info, you can simplify it using the base classes I added. Also val is always `DBString` so we don't need T, unless you want to support `std::string` now


```c++
template <typename U>
void processCommonEvent(const DBString&, const lcf::ContextStructBase<U>&) {}

void processCommonEvent(const DBString& val, const lcf::ContextStructBase<lcf::rpg::EventCommand>& ctx) {
 	// is rpg::EventCommand and compiles
        int indent = ctx.obj->indent;
} 

lcf::rpg::ForEachString(lcf::Data::data, [&](auto& val, auto& ctx) {
    processCommonEvent(val, ctx);
}
```

fdelapena (approved):

Ghabry:
Imo this is 99.9% perfect now. Multiple tools show that the API is good enough for complex processing tasks. (just needs some squashing in the end)

The only strange behaviour imo is that e.g.
```cpp
template <typename T> bool isCommonEventCommand(const lcf::ContextStructBase<T>&) { return false; }
bool isCommonEventCommand(
  const lcf::Context<lcf::rpg::EventCommand, lcf::ContextStructBase<lcf::rpg::CommonEvent>>& ctx)
{ return true; }
```
does not return ``true`` for event commands that belong to common events.

fmatthew5876:
Squashed, I think this PR is ready to go

PR #397, opened 2020-09-15
CMake: Fix expat, disable benchmarks

Ghabry:
Is a draft - Tests pending:

- [x] Builds on Windows
- [x] The new config file works


Ghabry:
just noticed that lcfstrings is missing in Makefile.am. Will also add this entry here.

CMakeLists.txt
```diff
@@ -442,12 +443,14 @@ add_dependencies(check_lcf test_runner_lcf)
 add_dependencies(check check_lcf)
 
 # benchmarks
-file(GLOB BENCH_FILES ${CMAKE_CURRENT_SOURCE_DIR}/bench/*.cpp)
-foreach(i ${BENCH_FILES})
-	get_filename_component(name "${i}" NAME_WE)
-	add_executable(bench_${name} ${i})
-	target_link_libraries(bench_${name} lcf)
-endforeach()
+if(NOT DISABLE_BENCHMARKS)
```
fmatthew5876, line 452: Should we call this `LIBLCF_ENABLE_BENCHMARKS` to match player?

Ghabry: I consider changing all the options to ``LCF_SOMETHING`` like we have in Player and deprecating all the ``DISABLE`` options.

Ghabry:
will also add the doctest WIN32 console fix here

CMakeLists.txt
```diff
@@ -1,13 +1,14 @@
 cmake_minimum_required(VERSION 3.7)
-cmake_policy(SET CMP0048 NEW)
+cmake_policy(SET CMP0091 NEW) # Support MSVC_RUNTIME_LIBRARY
```
hartwork, line 2: This policy is introduced in CMake 3.15 so my vote for a bump from 3.7 to 3.15 above or for a version check guard.

CMakeLists.txt
```diff
@@ -349,8 +350,14 @@ endif()
...
-	find_package(EXPAT)
-	target_link_libraries(lcf EXPAT::EXPAT)
+	find_package(expat)
+	if(NOT expat_FOUND)
+		# Fallback to old expat detection
+		find_package(EXPAT REQUIRED)
+		target_link_libraries(lcf EXPAT::EXPAT)
+	else()
+		target_link_libraries(lcf expat::expat)
+	endif()
```
hartwork, line 360: Flipping the order of branches would reduce cognitive load of this code by a bit. Same again further down one more time.

hartwork (commented):
Two small things I spotted, below:

CMakeLists.txt
Ghabry: Besides this feature on Windows we don't use anything from CMake past 3.7.

There is a version check in ``ConfigureWindows.cmake`` (and a noisy warning) already that they should update to CMake 3.15+ on Windows. Afaik CMake does not warn about unknown policies so an if-endif here would add 2 more unnecessary lines.

hartwork: I don't agree but your call :smiley:

Ghabry: There is really a if check required around the policy... otherwise this throws an error xD

Ghabry:
Jenkins: test this please

Ghabry:
There is still something fishy with the expat on Windows, will test this on Windows system later.

CMakeLists.txt
```diff
@@ -1,13 +1,30 @@
...
+		list(FIND old_params ${x} index)
+		list(GET new_params ${index} new_value)
+		message(FATAL_ERROR "${x} was renamed to ${new_value} and the logic inverted.")
+	endif()
+endforeach()
+
+option(BUILD_SHARED_LIBS "Build shared library, disable for building the static library (default: ON)" ON)
+option(LIBLCF_WITH_ICU "ICU encoding detection (when OFF fallback to iconv, not recommended, default: ON)" ON)
+option(LIBLCF_WITH_XML "XML reading support (expat, default: ON)" ON)
+option(LIBLCF_UPDATE_MIMEDB "Do not run update-mime-database after install (default: ON)" ON)
```
carstene1ns, line 24: Wrong description.

Ghabry:
Works now for me. Ready for review.

carstene1ns (approved):
Looks good to me.

fmatthew5876 (approved):

PR #403, opened 2020-09-26
Remove SaveMapInfo Setup and Fixup

fmatthew5876:
Depends on: https://github.com/EasyRPG/Player/pull/2358


fdelapena (approved):

generator/csv/fields.csv
```diff
@@ -985,8 +985,8 @@ SaveMapInfo,position_y,f,Int32,0x02,0,0,0,int
 SaveMapInfo,encounter_rate,f,Int32,0x03,-1,0,0,int
 SaveMapInfo,chipset_id,f,Int32,0x05,-1,0,0,int
 SaveMapInfo,events,f,Array<SaveMapEvent>,0x0B,,1,0,? array
-SaveMapInfo,lower_tiles,f,Vector<UInt8>,0x15,,1,0,? [00 01 02 ... 8E 8F]
-SaveMapInfo,upper_tiles,f,Vector<UInt8>,0x16,,1,0,
+SaveMapInfo,lower_tiles,f,Vector<UInt8>,0x15,"[x for x in range(0, 144)]",1,0,? [00 01 02 ... 8E 8F]
+SaveMapInfo,upper_tiles,f,Vector<UInt8>,0x16,"[x for x in range(0, 144)]",1,0,
```
Ghabry, line 989: May I suggest a small change to generate.py?

```py
        try:
            ilist = eval(dfl)
            iterator = iter(ilist)
            return " = {" + ', '.join(str(x) for x in iterator) + "}"
         except Exception as e:
             pass
```

Now you can write it as ``range(0, 144)`` because anything that is iterable is accepted.

fmatthew5876: I like the idea, but how do we differentiate python expressions from constants? Suppose I had an array constant I wanted to use?

The way it works now, the semantics are if the value is a python list expression, it uses `eval`, otherwise it will just copy the value verbatim.

Ghabry: Valid point, maybe the typecheck could be extended to whatever a range is: https://stackoverflow.com/questions/13092267/if-range-is-a-generator-in-python-3-3-why-can-i-not-call-next-on-a-range
But I don't want to investigate this right now, so lets merge this.

PR #406, opened 2020-10-14
Rename and add missing CBA chunks / structs

fmatthew5876:
Depended by https://github.com/EasyRPG/Player/pull/2407

DynRPG functions
--------------

| Address | Function |
| -- | -- |
| 00481F40 | RPG_Animation2::InitSaveChunk |
| 00481A14 | RPG_Animation2Pose::InitSaveChunk |
| 00481BCC | RPG_Animation2Weapon::InitSaveChunk |
| 0047F530 | RPG_Animation2ItemSkill::InitSaveChunk  |
| 0047DF28 | RPG_Item::Create |
| 0047EAB4 | RPG_Skill::Create |


fdelapena (approved):

fmatthew5876:
Reviewed this against DynRPG, all the structs, chunks, and chunk defaults confirmed.

No idea what `BattlerAnimationItemSkill::unknown02` is for. As far as I can see, it's not used at all by `RPG_RT` but it is written and read from the save data.

Maybe an unused feature?

Ghabry:
I didn't see a warning when I checked read/write years ago. So I can conclude that _every single 2k3 game_ on rmarchiv used the default CBA options 😂😂😂

Ghabry:
Wait a moment. We warn when parsing unknown chunks. Just in the terminal but I would have seen this in the last 8 years.
So this means the default database doesn't use this and also no game I played????? W T F

fmatthew5876:
I am just as confused as you are.

I can't believe no game ever tripped one of these errors in the past but it must be true. We did have most of the chunks before..

Ghabry:
Could you do me a favor?
Can you make a screenshot of the CBA Item Editor (I guess all missing chunks are from that editor?) and mark all fields that were missing with a MSPaint-Cross?
Want to see which features were never used in a visual way. :)

fmatthew5876:
Ok this makes sense

So looking at it again, the only real chunk "added" here is  `BattlerAnimationItemSkill::unknonwn02` which I'm not even sure if it appears at all in the UI.

* `BattlerAnimationExtension` was copied to `BattlerAnimationPose` - no chunks chaged
* `BattlerAnimationExtension` was copied to `BattlerAnimationWeapon` - 2 chunks removed
* `ItemAnimation` and `BattlerAnimationData` were combined into `BattlerAnimationItemSkill`. 

The last case is the most interesting. RPG_RT binary uses the same class for both of these, but the editor only lets you change different fields for items and skills.

For example in a skill you can change the pose so your skill could look like a weapon attack, or a skill use, item use, or something else etc.. So while `pose` is technically available for items, you can change it in the editor.

Maybe we should keep them as 2 separate classes. I'll check tonight whether or not skills let you use the uneditable weapon chunks or if weapons let you use the uneditable skill chunks

fdelapena (approved):

PR #407, opened 2020-10-15
rpg::Actor Add unknown 3A and 40 chunks found in DynRPG

fmatthew5876:


fdelapena (approved):

Ghabry:
btw while I was investigating the dungeon generator I'm pretty sure that I also found some unknown chunks in the code that reads the map properties:

In function ``487A34`` the values: 2B, 2C, 2D, 2E, 2F.

So I'm not really sure if we really want to document them all. There seem to be alot??

fmatthew5876:
I'm on the fence about this one.

If they are truly not used at all by RPG_RT, then there isn't much value in adding memory for them to our structures. That includes the chunks in this PR.

That being said, we do want to be sure the chunk IDs don't get reused for some custom feature later.

Ghabry:
Not reusing the chunks is a valid concern. Lets keep this open longer, bet it is not hard to find even more...

fdela suggested they could be editor-only chunks (temporary data that is not saved).
But this investigating the editor which is some work...

PR #409, opened 2020-10-17
Remove lcf::Data

### LINKED ISSUE #305: Wrong event page variable condition when writing emu to lmu
RPG Maker version: 2003 1.12a
Lastest nightly liblcf in LCF2XML (Jan24 2019).

When you convert a map with a event that contains a tab with a variable condition from EMU to LMU (LMU to EMU is fine), the condition is changed.

Here's a test case:
We have this condition in a map.
image

Then after converting the map from LMU to EMU, and then back to LMU with LCF2XML, the condition changes to this one:
image

fmatthew5876:
Depended by: https://github.com/EasyRPG/Player/pull/2404
Depended by: https://github.com/EasyRPG/Editor-Qt/pull/150

Fix #305


fdelapena (approved):

Ghabry (commented):
How about making SaveOpt a non-optional param and adding 2k3 as a SaveOpt?

fmatthew5876:
I'm on the fence about that. I kind of feel this flag is so important it should stand on it's own so you're forced to think about it and set the right value. Also it's not used for `LDB_Reader`.

I could be convinced otherwise though

Ghabry:
Instead of only adding Engine2k3 one could also make the functions hard fail (assert?) when the ``SaveOpt`` is not set.

Add ``Engine2k`` and ``Engine2k3`` to it and check for the existence.

If you don't like this: I would prefer, instead of the ``2k3 bool`` to have an engine enum here. Then this can be extended if we decide to to e.g. only write fields when the engine is ``EasyRpg``.

fmatthew5876:
Changed the bool to an enum, I agree this is better.

Once this PR is approved, I will update Player and Editor PRs

Ghabry:
on first glance this looks fine. Reviewing it tomorrow, thanks.

fdelapena (approved):

Ghabry (approved):
waiting for the other PR updates before merging this

fmatthew5876:
All PRs updated, this is ready now

PR #416, opened 2021-03-15
Allow redirection of lcf console messages

Ghabry:
First commit: The chunk dumping during Skip didnt really work except on Windows. I finally fixed it (useful for Maniacs Patch to find unknown chunks). Also is now always on, a missing chunk is now always a bug because we know about all common ones.

Second commit: **I want your feedback here**. Before I implement this completely (is some work) please tell me if the general approach is fine for you.
I want to allow custom output handlers for lcf diagnostics. E.g. for the Player they could be written to easyrpg_log.txt. Currently they are completely lost which is bad for handling bug reports. As vital information is missing.

To redirect it library users can invoke lcf::LogHandler::SetHandler(_Handlerfunctionhere_)

### DIFF
--- a/src/lcf/log_handler.h
+++ b/src/lcf/log_handler.h
@@ -1,0 +1,47 @@
+/*
+ * This file is part of liblcf. Copyright (c) 2020 liblcf authors.
+ * https://github.com/EasyRPG/liblcf - https://easyrpg.org
+ *
+ * liblcf is Free/Libre Open Source Software, released under the MIT License.
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
+#ifndef LCF_OUTPUT_H
+#define LCF_OUTPUT_H
+
+#include "string_view.h"
+
+namespace lcf {
+namespace LogHandler {
+
+enum class Level {
+	Debug,
+	Warning,
+	Error,
+	Highest
+};
+
+using LogHandlerFn = void (*)(Level level, StringView message);
+
+/**
+ * Sets the output handler for all lcf logging.
+ * The default handler prints to standard error.
+ *
+ * @param fn New output handler. nullptr for default handler.
+ */
+void SetHandler(LogHandlerFn fn);
+
+/**
+ * Only report issues that have at least this log level.
+ * Use Highest to disable logging.
+ * Default: Debug
+ *
+ * @param new_level New log level
+ */
+void SetLevel(Level new_level);
+
+} // namespace LogHandler
+} // namespace lcf
+
+#endif
--- a/src/lcf/log_handler.h
+++ b/src/lcf/log_handler.h
@@ -1,5 +1,5 @@
 /*
- * This file is part of liblcf. Copyright (c) 2020 liblcf authors.
+ * This file is part of liblcf. Copyright (c) liblcf authors.
  * https://github.com/EasyRPG/liblcf - https://easyrpg.org
  *
  * liblcf is Free/Libre Open Source Software, released under the MIT License.
--- a/src/lcf/log_handler.h
+++ b/src/lcf/log_handler.h
@@ -11,6 +11,7 @@
 #define LCF_OUTPUT_H
 
 #include "string_view.h"
+#include "enum_tags.h"
 
 namespace lcf {
 namespace LogHandler {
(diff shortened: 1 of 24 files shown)


fdelapena:
There's a good chance to add more levels, such `trace`, `info`, `notice`, while `critical`, `alert`, `emergency` feel a bit unneeded.
At least those for matching not just with Player, but cover at least the levels existing in JavaScript's `console.xxx()`. Emscripten shell currently uses `console.error()` for everything but this is easy to fix and helps a lot to filter details from the browser developer tools.

aleck099:
Why is exceptions disabled in liblcf?

Ghabry:
Same reason why we disable RTTI:

Portability. We have some quite low spec homebrew platforms that we support.
They are low on memory and load the executable into memory so every kilobyte counts.
Also exceptions are also not well supported in emscripten.

We also do not want to maintain two codepaths. So no exception it is.

aleck099:
But is printing to stderr a good choice?

Ghabry:
Uhm no that is why I plan to make this configurable in this PR?

Currently errors land on stderr.
With this PR consumers of this library (e.g. the Player) can hook it into their logging system (which logs into easyrpg_log.txt)

aleck099:
Looking forward to begin merged

Ghabry:
I rechecked the code now: There is literally nothing finished here yet. Except for one example usage. Can't finish this for 0.8.0 in such short time.

Hot candidate for 0.8.1

carstene1ns:
I think on liblcf side is not much missing anymore. The handler function concept is well-known and works. Usually you get some `void *userdata` context to pass, which we maybe should make available for custom loggers.

Ghabry:
Suitable logging function for use in Player:

```cpp
lcf::LogHandler::SetHandler([](lcf::LogHandler::Level level, StringView message, lcf::LogHandler::UserData) {
    Output::Debug("lcf ({}): {}", lcf::LogHandler::kLevelTags.tag(level), message);
});
```

So currently only written to terminal/logfile.

Not sure if any of the stuff should be reported through the graphical "console" inside the window. Most of that stuff is recoverable (similar to the libpng warnings spamming the shell) and everything critical is already shown as an "Output::Error".

src/lcf/log_handler.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * This file is part of liblcf. Copyright (c) 2020 liblcf authors.
```
carstene1ns, line 2: copyright years here and in other `log*` files

carstene1ns (approved):
I find it a bit messy that at first every log message is made a C-String, then a C++ std::string and last a string_view.
Maybe we should make `fmt` a lcf dependency in the future to hide the conversions.

Apart from the minor nits, LGTM

src/log.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+#ifndef LCF_LOG_H
+#define LCF_LOG_H
+
+#include "lcf/log_handler.h"
+
+namespace lcf {
+namespace Log {
+
```
carstene1ns, line 17: One thing that could be added here is argument checking by `g++`
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute

I have hidden this behind a preprocessor define here:
```c++
#ifdef __GNUG__
	#define LIKE_PRINTF __attribute__((format(printf, 1, 2)))
#else
	#define LIKE_PRINTF
#endif
```

Ghabry: done and it even found a bug :D

fdelapena (approved):

src/log.h
```diff
@@ -12,12 +12,18 @@
...
 
-void Debug(const char* fmt, ...);
-void Warning(const char* fmt, ...);
-void Error(const char* fmt, ...);
+void Debug(const char* fmt, ...) LIKE_PRINTF;
+void Warning(const char* fmt, ...) LIKE_PRINTF;
+void Error(const char* fmt, ...) LIKE_PRINTF;
 
 } // namespace Log
 } // namespace lcf
```
carstene1ns, line 29: Minor nit: we should `#undef LIKE_PRINTF` here to not pollute other projects with internal stuff.

carstene1ns (approved):

src/reader_lcf.cpp
```diff
@@ -276,21 +279,27 @@ int LcfReader::Peek() {
...
+	ss << std::hex;
+
 	for (uint32_t i = 0; i < chunk_info.length; ++i) {
 		uint8_t byte;
 		LcfReader::Read(byte);
-		fprintf(stderr, "%02X ", byte);
+		ss << std::setfill('0') << std::setw(2) << (int)byte << " ";
 		if ((i+1) % 16 == 0) {
-			fprintf(stderr, "\n");
+			Log::Debug(ss.str().c_str());
```
carstene1ns, line 293: Android (clang) build fails with `-Wformat-security`:

```c++
Log::Debug("%s", ss.str().c_str());
```

PR #420, opened 2021-04-02
Make headers private sources

carstene1ns:
This slipped in with #361.

>From https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/
>
>The PRIVATE and PUBLIC keywords specify where those corresponding sources should be used. PRIVATE simply means those sources should only be added to myLib, whereas PUBLIC means those sources should be added to myLib and to any target that links to myLib. An INTERFACE keyword can be used for sources that should not be added to myLib but should be added to anything that links to myLib. In practice, sources will almost always be PRIVATE, since they shouldn’t generally be added to anything that links against the target. Header-only interface libraries are one exception because sources can only be added as INTERFACE for interface libraries. Do not confuse the PRIVATE , PUBLIC and INTERFACE keywords with whether a header is part of the public API for the library or not, the keywords are specifically for controlling which target(s) the sources are added to in this case.


fdelapena (approved):

carstene1ns:
test this please

Ghabry (approved):

PR #422, opened 2021-05-14
Support CMake 3.10-3.20

carstene1ns:
Add github action to check for breakage.


CMakeLists.txt
```diff
@@ -1,23 +1,23 @@
-cmake_minimum_required(VERSION 3.7)
-
-if(POLICY CMP0091)
-	cmake_policy(SET CMP0091 NEW) # Support MSVC_RUNTIME_LIBRARY
+cmake_minimum_required(VERSION 3.10...3.20 FATAL_ERROR)
+if(${CMAKE_VERSION} VERSION_LESS 3.12)
+	# Use most recent policies, even for older cmake
+	cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
```
carstene1ns, line 4: The above is to specify "supported versions" and to enable all policies.
(See https://cliutils.gitlab.io/modern-cmake/chapters/basics.html)

Ghabry:
- The module ``FindEXPAT`` was added in 3.10 and can be removed
- The module ``FindICU`` was added in 3.7 and can be removed. There is an improvement in 3.11 for Windows but on Windows we usually have the version VIsual Studio bundles, so one of the latest
- ``FindIconv`` was added in 3.11, so must stay (for now ;))

.github/workflows/stable-compilation.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          export DEBIAN_FRONTEND="noninteractive"
+          apt-get update
+          apt-get install -y --no-install-recommends --no-install-suggests \
+            build-essential cmake ninja-build git \
+            libicu-dev libexpat1-dev
```
Ghabry, line 41: This is also great for an updated "how to install" guide because I see the package names to install :D

Ghabry (approved):
Modules must be deleted, otherwise: LGTM

.github/workflows/stable-compilation.yml
carstene1ns: Debian still has funny names for some libs. For player it is worse.

fdelapena (approved):

CMakeLists.txt
```diff
@@ -309,11 +296,7 @@ set(LCF_HEADERS
 	src/lcf/third_party/string_view.h
 )
 
-list(TRANSFORM LCF_SOURCES PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(TRANSFORM LCF_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
-list(APPEND LCF_HEADERS ${CMAKE_CURRENT_BINARY_DIR}/src/lcf/config.h)
```
carstene1ns, line 314: cc #423.

PR #426, opened 2021-05-31
Ignore /.github/ to prevent affecting forks

fdelapena:
This was working fine so far with issue templates.
However, 2FA accounts need additional personal token permissions otherwise push gets rejected when trying to upload workflow files. To prevent discouraging contributions from forks due to the additional maintenance burden, ignoring this folder seems safe.

**Note:** the 2FA issue is only reproducible when the fork adds or modifies new files to `/.github/workflows/` and then trying to push them to the fork.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ Makefile
 
 # tests
 test_runner*
+
+# GitHub
+/.github/


carstene1ns:
Are you using some other app besides git command line?

Ghabry:
I don't really understand when this is a problem. Can you give an example?
Does GitHub document this anywhere?

carstene1ns:
I only found this is a problem when using github desktop,vscode or some other tools with old tokens.

fdelapena:
I've been hit by this issue when the following conditions are met:

- Having 2FA enabled to login, so I was using a personal token (it's an old token) for the password. The token does not have the `workflow` scope set.
- Having an older fork not containing the `workflows/stable-compilation.yml` file
- Tried from command line `git pull upstream master` then `git push origin master` to sync my fork

git failed to push with:
```
! [remote rejected] master -> master (refusing to allow a Personal Access Token to create
or update workflow `.github/workflows/stable-compilation.yml` without `workflow` scope)
```

This is not reproducible when at least one of the following are met:
- The fork and git clone was modern enough to contain the file so not needing to add or update workflow files (tested)
- The personal token gets updated or set with that permission

PR #431, opened 2021-08-27
Remove the need for pkgconfig in cmake config

carstene1ns:
- Untested on windows and macos
- Fixes using of liblcf with psvita toolchain
- Simplifies finding of the library

Fun fact: most of the autotools hackery is only needed for distributions like debian and mageia (hello #341), that put stuff in custom directories.


Ghabry:
confirmed working on macOS

Ghabry (approved):
works on Windows

PR #434, opened 2021-09-01
SavePicture: Add chunks for Flip and Blend Mode

Ghabry:
Contrary to KeyInput where the usage of the upper bits made sense here it is far too ugly to handle this way:

Maniac Patch uses the upper bits of finish_bot_trans for this which is imo very very ugly.

Instead I add here 3 new chunks, so savegames will be incompatible.

The Blend Mode is an integer to get more flexibility compared to Multiply/Add/Overlay of Maniacs.


fdelapena (approved):

carstene1ns:
IMO the flip mode should be an enum: `FlipNone`, `FlipX`, `FlipY`, `FlipBoth`, it would need only one chunk this way.
However, two chunks is also okay.

Ghabry:
reduced to one Flip chunk and added a new enum for the "effect": This is for the fixed angle (instead of rotating)

carstene1ns (approved):

PR #435, opened 2021-09-07
Improve encoding detection

Ghabry:
The first commit is not really needed for it but it makes it after 0.7 easier to replace StringView with std::string_view (C++17). This is API compatible, no changes needed to consumers.

The second commit changes the Encoding Api from Streams to a Database-handle. Because we do not operate on global objects anymore the database checked by player was always empty. This fixes it (Player loads the DB and passes it in). I also noticed that for whatever reason the language detection is better when system tab is at the beginning (maybe filenames are better than terms because they are usually unaffected by translations?)

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -75,8 +75,8 @@ std::string ReaderUtil::CodepageToEncoding(int codepage) {
 	return outs;
 }
 
-std::string ReaderUtil::DetectEncoding(std::istream& filestream) {
-	std::vector<std::string> encodings = DetectEncodings(filestream);
+std::string ReaderUtil::DetectEncoding(lcf::rpg::Database& db) {
+	std::vector<std::string> encodings = DetectEncodings(db);
 
 	if (encodings.empty()) {
 		return "";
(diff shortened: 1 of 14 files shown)


fdelapena (approved):

src/reader_util.cpp
```diff
@@ -85,114 +85,111 @@ std::string ReaderUtil::DetectEncoding(std::istream& filestream) {
...
+		});
+		return ascii ? std::string() : (ToString(s) + " ");
+	};
+
+	lcf::rpg::ForEachString(db.system, [&](const auto& val, const auto& ctx) {
+		text << filter(val);
+	});
+
+	// Cannot use ForEachString here for Terms:
+	// Too much untranslated garbage data in their, even in default database
```
carstene1ns, line 104: typo: "in there"

carstene1ns (commented):
Question is, if the encoding is not detected for ascii strings anymore (filtered out), how to know it is an english game at all?

Ghabry:
Thats a good question. ICU seems to detect UTF-8 in that case. This may not be ideal, usually you want western european in that case...

Ghabry:
Checked now the ICU sourcecode: They use ngrams (Split string in pairs of N chars) for language detection (this is an approach that works surprisingly good btw) so removing ASCII strings will actually make the result _worse_ as it messes up the distribution. I will remove it and test again.

(For the ZIP archive encoding detection were I took this from this still makes sense though as _all_ files are known beforehand, so ASCII implies UTF-8)

Ghabry:
rechecked this: using the system tab before terms is good enough. Ascii filter was nonsense :)

fdelapena:
Related (fixes?): #169

Ghabry:
it is still read twice. Not possible until the save data is also using DB String. Maybe 0.7.1 ;)

carstene1ns (approved):

carstene1ns:
lcftrans is broken now :P

Ghabry:
yeah, everything that does encoding detection needs the API updated. Will provide a fix soon

PR #445, opened 2022-12-07
Add custom chunks required for ShowStringPicture

Ghabry:
This is implemented with a new data structure EasyRpgWindow that contains EasyRpgText.

When we ever add our own Window event functions these will provide a good baseline.

### Design

The window has the same ID as the picture. The picture receives a new field ``easyrpg_type`` to indicate if anything is attached to the picture.

The savegame receives a new chunk "2" to the EasyRPG data: ``Array<SaveEasyRpgWindow>``.

A ``SaveEasyRpgWindow`` describes a single window. Right now only ShowStringPic can create them.

A Window has styling information and a list of texts (``SaveEasyRpgText``).

Window styling stuff: width, height, system_name, message_stretch, flags like "draw_frame".

``SaveEasyRpgText`` contains stuff like "text" (obviously), x, y, font, font size, letter spacing (unsupported), line spacing, flags like italic or bold.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,10 @@ set(LCF_SOURCES
 	src/generated/lsd_saveactor.cpp
 	src/generated/lsd_savecommonevent.cpp
 	src/generated/lsd_saveeasyrpgdata.cpp
+	src/generated/lsd_saveeasyrpgtext.cpp
+	src/generated/lsd_saveeasyrpgtext_flags.h
+	src/generated/lsd_saveeasyrpgwindow.cpp
+	src/generated/lsd_saveeasyrpgwindow_flags.h
 	src/generated/lsd_saveeventexecframe.cpp
 	src/generated/lsd_saveeventexecstate.cpp
 	src/generated/lsd_saveinventory.cpp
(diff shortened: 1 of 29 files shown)


CMakeLists.txt
```diff
@@ -187,6 +193,7 @@ set(LCF_SOURCES
 )
 
 set(LCF_HEADERS
+	src/lcf/config.h
```
Ghabry, line 196: wups, this was added by accident 🤔

jetrotal:
Unrelated but found while testing:
Audio fails when `pitch = v[n]` 
It goes to 0%, or something like this


EasyRPG Dev Simulator project to test it:
[Project2.zip](https://github.com/EasyRPG/liblcf/files/10188776/Project2.zip)

Ghabry:
The v[n] for sound effects must be a custom Maniac thing. The original does not support indirection. Though this is very likely super simple to implement. Please report it with parameters in the maniac meta bug

MackValentine:
Hi !
I found 2 bug with StringPic :
- the origin is not taken into consideration
To fix it I just add "ApplyOrigin(false);" inside the "AttachWindow" function

- If you use a StringPic, then a normal Picture with the same ID, the Picture will keep the StringPic inside the "new" Picture
To fix it I just add "data.easyrpg_type = lcf::rpg::SavePicture::EasyRpgType_default;" at the beginning of the "Show" function

( Sorry I'm really a noob with Git, I don't know how I can explain better >< )

Ghabry:
just found a bug in liblcf: Flags with "default values" were never unset.

I wondered why StringPics to not preserve the flags correctly.

This also affects pictures: A picture with flag "remove on map change" disabled will have the flag enabled after loading. upps.

fdelapena (approved):

PR #451, opened 2023-01-22
CMake config: 3rd try

Ghabry:
Tested now the following combinations:

- liblcf build with autotools
- liblcf build with CMake

For both static and shared.

Then compiled Player:

- CMake with liblcf autotools: Both "pkg-config" and "pkg-config --static" tested (4 builds)
- CMake with liblcf CMake (2 builds)

Everything works. This stuff is so dumb.


fdelapena (approved):

carstene1ns (approved):
lgtm

PR #457, opened 2023-04-26
Release 0.8

Ghabry:


Ghabry:
I attempted to build the resulting tarball using autotools and CMake and Player is also able to be built from it. So it should good to go.

fdelapena (approved):

PR #458, opened 2023-05-04
Require C++17

Ghabry:
plus removing legacy stuff that is hopefully solved by now by Microsoft...


Ghabry:
okay forget about removing old workarounds... they are still having the same bugs in latest VS2022 compiler m(

fdelapena (approved):

PR #464, opened 2023-07-25
CMake: Fix Big endian detection when crosscompiling

Ghabry:
Needs at least CMake 3.20 to work properly

(untested)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -322,7 +322,11 @@ if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
 endif()
 
 # endianess checking
-if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.20)
+	if (CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
+		target_compile_definitions(lcf PRIVATE WORDS_BIGENDIAN=1)
+	endif()
+else()
 	include(TestBigEndian)
 	test_big_endian(WORDS_BIGENDIAN)
 	if(WORDS_BIGENDIAN)


CMakeLists.txt
```diff
@@ -322,7 +322,11 @@ if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
 endif()
 
 # endianess checking
-if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.20)
+	if (CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
+		target_compile_definitions(lcf PRIVATE WORDS_BIGENDIAN=1)
+	endif()
+elseif(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
```
carstene1ns, line 329: Let's make this `else()`

The test binary is not executed, just checked for strings. So this is cheap enough.

carstene1ns (changes requested):

CMakeLists.txt
Ghabry: Actually an unconditional ``test_big_endian(WORDS_BIGENDIAN)`` would also work for >= 3.20 because they updated the script to use the new variables.

I will change it to this if-else here, then we can remove the else in 10 years when we update CMake again xD

carstene1ns (approved):
Works as expected.

CMakeLists.txt
carstene1ns: Next station is 3.18¹, so the road to 3.20 is not long anymore :D
I think we can skip 3.16², or wait until April '25.

1: https://packages.debian.org/search?keywords=cmake&searchon=names&exact=1&suite=all&section=all
2: https://packages.ubuntu.com/search?keywords=cmake&searchon=names&exact=1&suite=all&section=all

PR #467, opened 2023-09-26
add lcf2xml python build

seokjin1013:
Hello

I've made some modifications to enable the use of this library in Python. The build process remains unchanged, and only need to have Python installed.

I've wanted to use this library in a Firebase Python server environment, where file I/O is restricted. I think this may some helpful for someone.

```python
import pylcf2xml
ldb = open('RPG_RT.ldb', 'rb').read()
xml = pylcf2xml.ldb2xml(ldb, encoding='949')
print(xml)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ build/
 # Python generated files
 *.pyc
 __pycache__/
+python/venv
 
 # other generated files
 Makefile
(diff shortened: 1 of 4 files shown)


carstene1ns:
pybind11 docs say this is a good way:
```cmake
find_package(Python 3.6 COMPONENTS Interpreter Development REQUIRED)
find_package(pybind11 CONFIG REQUIRED)
pybind11_add_module(example example.cpp)
```

No need to use a virtual environment, just use the system package.

PR #470, opened 2023-11-22
EnumTags: Support non-monotonic enums (e.g. EventCommands)

Ghabry:
Asked for by @jetrotal to allow a simple way for Event Command lookup by String without creating a seperate lookup in Player.

Please do not look at ``enum_tags.h``. The template foo is difficult. Not an expert in this part of C++. I read various sources to figure some of the complex stuff, e.g. how to allow calling the constructor with "int, string, int string, ..." and storing it into the array.

No ChatGPT code involved, it isn't very good in advanced template programming :sweat_smile: (typical ChatGPT issue again: For the simple stuff I do not need it and for the hard stuff it is not helpful :sob: )

The hard part for me (Maybe should learn a bit about parameter packs ``...args`` then this is likely easier to do :D)

```cpp
template<typename... Args>
explicit constexpr EnumTags(Args const&...args) noexcept {
    AddTag<0>(args...);

    int_type i = 0;
    for (const auto& it : _tags) {
        if (it.value != i) {
            monotonic_from_zero = false;
        }
        ++i;
    }
}

template<std::size_t I, typename... Next>
constexpr void AddTag(E value, const char* name, Next const&...next) noexcept {
    _tags[I] = {int_type(value), name};

    if constexpr (sizeof...(Next) > 0) {
        AddTag<I + 1>(next...);
    }
}
```

------

**Do not merge. breaks Player. Needs a patch first!**


Ghabry:
readded ``makeEnumTags`` to make the code in the typical case (monotonic) less verbose. This also is now backwards compatible and wont break Player

fdelapena (approved):

PR #474, opened 2023-12-31
Update workflow and cmakery

carstene1ns:
- Better use as submodule, without install target (useful for flatpak)
- Prevent in-source builds


Ghabry (approved):
sorry for ignoring such an easy-to-review change for so long...

I fixed the runners (``ca-certificates`` added) and the Windows PR builder (in source build).

fdelapena (approved):

Ghabry:
The Player flatpak file requires an update because it does an in-source build of liblcf

PR #477, opened 2024-03-28
Fix building for Android

carstene1ns:
Resolve int warning (mostly on 3DS)


Ghabry:
I didn't knew before yesterday that we have no pr builder for Android here 😅

PR #479, opened 2024-04-14
Do not vendor inih

Ghabry:
The C++ part of the library is still kept as we made changes to it.

One patch less for Fedora ;)


carstene1ns (approved):
Looks okay to me.

~~Have installed `inih-devel` on server now to solve jenkins failure.~~ EasyRPG/buildscripts#174
Let`s just hope the API does not change so our vendored part will not break.

carstene1ns:
jenkins: test this please

Ghabry:
all green. only flatpak needs inih added.

PR #484, opened 2024-09-27
Change default CE to have the "call" trigger type

lumiscosity:
Fixes CEs initializing with an invalid trigger type value of 0 by properly replacing it with 5 ("call" type).


carstene1ns:
jenkins: ok to test

fdelapena (approved):

Ghabry (approved):
right, this also matches with the value in the enum.

Went unnoticed, because:

- Player code only checks for autorun and parallel
- Default values are not written into the lcf files, so the wrong value never appears inside the files (so not even testing in RPG_RT.exe could catch this)

PR #486, opened 2024-10-10
Support ICU library provided by Windows 10 + Preset updates

Ghabry:
Ported the normalization code to the ICU C-API so looks more ugly now :sweat_smile: but C++ has no stable ABI so Microsoft decided to only provide the C part.

This _does not_ remove 3rd party ICU support. When a custom ICU library is found it has preference.

Rest is just updating of presets based on Player preset updates.

---

For now I plan to keep using the vcpkg ICU library but when I do not want to maintain this anymore I can now simply drop it.

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -10,12 +10,16 @@
 #include "lcf/config.h"
 #include "lcf/scope_guard.h"
 
-#if LCF_SUPPORT_ICU
+#if LCF_SUPPORT_ICU == 1
 #   include <unicode/ucsdet.h>
 #   include <unicode/ucnv.h>
 #   include <unicode/normalizer2.h>
-#   include <unicode/unistr.h>
-#   include <unicode/locid.h>
+#   include <unicode/ustring.h>
+#elif LCF_SUPPORT_ICU == 2
+#	ifndef _WIN32
+#		error "icu.h only supported on Windows"
+#	endif
+#	include <icu.h>
 #endif
 
 #ifdef _WIN32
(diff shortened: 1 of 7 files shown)


fdelapena (approved):

carstene1ns:
Something is failing here with function renaming across multiple ICU versions.
Maybe some inclusion order of headers or something?

src/reader_util.cpp
```diff
@@ -10,12 +10,17 @@
 #include "lcf/config.h"
 #include "lcf/scope_guard.h"
 
-#if LCF_SUPPORT_ICU
+#if LCF_SUPPORT_ICU == 1
 #   include <unicode/ucsdet.h>
 #   include <unicode/ucnv.h>
 #   include <unicode/normalizer2.h>
 #   include <unicode/unistr.h>
 #   include <unicode/locid.h>
```
carstene1ns, line 18: `unistr.h` is c++ api

```suggestion
#   include <unicode/ustring.h>
```

src/reader_util.cpp
```diff
@@ -268,27 +273,74 @@ std::string ReaderUtil::Recode(StringView str_to_encode, StringView source_encod
...
+	}
+
 #if LCF_SUPPORT_ICU
-	icu::UnicodeString uni = icu::UnicodeString(str.data(), str.length(), "utf-8").toLower(icu::Locale::getRoot());
 	UErrorCode err = U_ZERO_ERROR;
-	std::string res;
-	const icu::Normalizer2* norm = icu::Normalizer2::getNFKCInstance(err);
+
+	auto log_warning = [err, &str](StringView message) {
+		Log::Error("%s failed while normalizing \"%s\": %s", std::string(str).c_str(), u_errorName(err));
```
carstene1ns, line 284:
```suggestion
	auto log_warning = [err, &str](const char *func_name) {
		Log::Error("%s failed while normalizing \"%s\": %s", func_name, std::string(str).c_str(), u_errorName(err));
```

carstene1ns (changes requested):

Ghabry:
thanks for the fixes c1, added them

carstene1ns (approved):

PR #490, opened 2025-01-03
Don't strictly require inih

Antidote:
inih isn't used by anything in liblcf directly, relaxing this requirement and allowing it to be disabled enables outside projects to more easily integrate this library into their application.

This PR enables this while still making inih a default requirement, so the current behavior is the default one requiring no changes in existing apps (easyrpg-player)

### DIFF
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -206,7 +206,7 @@ std::vector<std::string> ReaderUtil::DetectEncodings(StringView string) {
 	return encodings;
 }
 
-#if LCF_SUPPORT_INIH
+#if LCF_SUPPORT_INI
 std::string ReaderUtil::GetEncoding(StringView ini_file) {
 	INIReader ini(ToString(ini_file));
 	if (ini.ParseError() != -1) {
--- a/src/reader_util.cpp
+++ b/src/reader_util.cpp
@@ -215,6 +215,8 @@ std::string ReaderUtil::GetEncoding(StringView ini_file) {
 			return ReaderUtil::CodepageToEncoding(atoi(encoding.c_str()));
 		}
 	}
+#else
+	Log::Warning("Could not get encoding from ini file, disabled in this liblcf build.");
 #endif
 	return {};
 }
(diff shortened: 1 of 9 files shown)


Ghabry:
Thanks that looks like a reasonable change. Will look at it during the weekend.

fdelapena:
jenkins: ok to test

Ghabry (approved):

Ghabry:
Jenkins: test this please

Strange WiiU build error o_O ... lets try again.

Antidote:
Somehow I missed a few references and they cropped up when I started integrating it.

Antidote:
What the heck? I literally just built it on linux o.0

Antidote:
Well that was an ordeal, but everything should be in order now. lcf2xml was being naughty and threw off my assumption lol

Ghabry:
more junk for autotools was missing... should be fine now. :)

src/reader_util.cpp
```diff
@@ -207,24 +207,28 @@ std::vector<std::string> ReaderUtil::DetectEncodings(StringView string) {
...
 std::string ReaderUtil::GetEncoding(StringView ini_file) {
+#if LCF_SUPPORT_INI
 	INIReader ini(ToString(ini_file));
 	if (ini.ParseError() != -1) {
 		std::string encoding = ini.Get("EasyRPG", "Encoding", std::string());
 		if (!encoding.empty()) {
 			return ReaderUtil::CodepageToEncoding(atoi(encoding.c_str()));
 		}
 	}
+#endif
```
carstene1ns, line 218: Suggestion: Maybe add some message instead of failing silently?
```c++
#else
Log::Warning("Could not get encoding from ini file, disabled in this liblcf build.");
#endif
```

Antidote: Yeah should probably do that.

Ghabry:
Okay i think all are happy now. Dependency changes are always a bit annoying to code 😅

Antidote:
Yeah, having to deal with it in my own projects gives me that understanding as well lol.

PR #492, opened 2025-02-12
Flags: Do not hardcode one Flag field per class

Ghabry:
This is an alternative approach for #491

Instead of a single variable it uses our Flag feature to provide bool variables. This reduces the need for bitmasking.

Problem was that our code always hardcoded ``StructName_Flags`` for Flags. This restriction is now lifted and multiple Structs can have the same flags and a struct can have multiple flags.

_Don't merge, Player build failure because classes are renamed_


fdelapena (approved):

Ghabry:
@florianessl this should be okay now?

florianessl (approved):

Ghabry:
Jenkins: test this please

generator/csv/flags_easyrpg.csv
```diff
@@ -5,3 +5,25 @@ SaveEasyRpgText,draw_gradient,0
...
+EasyRpgStateRuntime,conf_override_active,0
+EasyRpgStateRuntime,reserved_1,0
+EasyRpgStateRuntime,reserved_2,0
+EasyRpgStateRuntime,reserved_3,0
+EasyRpgStateRuntime,patch_destiny_on,0
+EasyRpgStateRuntime,patch_destiny_off,0
+EasyRpgStateRuntime,patch_dynrpg_on,0
+EasyRpgStateRuntime,patch_dynrpg_off,0
+EasyRpgStateRuntime,patch_maniac_on,0
+EasyRpgStateRuntime,patch_maniac_of,0
```
florianessl, line 19: small typo here -> patch_maniac_of*

src/generated/lcf/rpg/saveeventexecstate.h
```diff
@@ -53,8 +54,47 @@ namespace rpg {
...
+					bool patch_keypatch_off;
+					bool patch_rpg2k3_cmds_on;
+					bool patch_rpg2k3_cmds_off;
+					bool use_rpg2k_battle_system_on;
+					bool use_rpg2k_battle_system_off;
+				};
+				std::array<bool, 20> flags;
+			};
+			EasyRpgStateRuntime_Flags() noexcept
+			{}
```
florianessl, line 84: Oh, another thing. Sorry for missing that when I first reviewed this:
These flag structures don't seem to be properly initialized. So whenever a flag field is accessed on a newly created SaveEventExecState, there is just garbage data here. :/

Maybe just append a line in the constructor here?
`flags.fill(false);`

Ghabry: Hu. I think you found a general bug applicable to all flags 😅.

Ghabry:
``a default member initializer is not allowed for a member of an anonymous struct within a union`` but only when using MSVC :cry:   okay so the old style then...

This wasn't an old bug but just something I missed when refactoring: The code was only executed when the flag field had the name STRUCT_FLAGS.

fdelapena (approved):

florianessl:
Small side effect: The renaming of the flag fields is also applied to generated XML files.
image

This also breaks the Players ability to read these fields from .emu files which were generated with previous versions of lcf2xml.

Ghabry:
Fixed.

Actually we do not guarantee a stable structure for the XML file across versions.

But I have to admit that this looks really ugly with the ``CLASS_CLASS_FLAG``.

PR #495, opened 2025-03-02
Migrate from string view lite to std::string_view

Ghabry:


fdelapena (approved):

fdelapena (approved):

src/inireader.cpp
```diff
@@ -100,41 +101,38 @@ int INIReader::ParseError() const
 	return _error;
 }
 
-std::string INIReader::Get(const std::string& section, const std::string& name, const std::string& default_value) const
+std::string_view INIReader::Get(std::string_view section, std::string_view name, std::string_view default_value) const
 {
 	std::string key = MakeKey(section, name);
 	// Use _values.find() here instead of _values.at() to support pre C++11 compilers
```
carstene1ns, line 107: Not our "bug", but we can safely migrate to `at()` here.

https://stackoverflow.com/a/70559609

carstene1ns (approved):
Maybe we could have gotten away with a `using StringView = std::string_view`, but of course this is cleaner. :D

Ghabry:
> Maybe we could have gotten away with a using StringView = std::string_view
> but of course this is cleaner. :D

We would have :D. Besides start and endwith the interface is the same ;)

PR #499, opened 2025-03-18
Add more Maniac Chunks and compile fix

Ghabry:


florianessl:
While we're still at it... This is pretty much a clear-cut case:

**structs_easyrpg.csv**

`ldb,StringVar,,1`

**fields_easyrpg.csv**
```
Database,maniac_stringvars,f,Array<StringVar>,0x21,,1,1,
StringVar,name,f,DBString,0x01,,0,0,String
```

Verified trough the content of latest tpc version (tcp_241028 -> see def_ldb.txt)

src/lcf/enum_tags.h
```diff
@@ -13,7 +13,6 @@
 #include <type_traits>
 #include <iterator>
 #include <cstddef>
-#include <array>
```
florianessl, line 16: By removing this line, Player compilation fails, due to std::array being undefined in `game_ineluki.h`. Looks like this was imported indirectly through some other header file before.
Make sure to add an #include to a Player PR if you haven't already.

Other than that, looks fine to me.

florianessl (approved):

src/lcf/enum_tags.h
Ghabry: good catch! I didn't test compiling because the lcf unit tests passed.

fdelapena (approved):

fdelapena (approved):

carstene1ns (approved):

PR #508, opened 2026-04-23
Fix Valgrind uninitialized memory warning when running tests

staticssleever668:
> Conditional jump or move depends on uninitialised value(s)

### DIFF
--- a/src/lcf/dbbitarray.h
+++ b/src/lcf/dbbitarray.h
@@ -129,7 +129,7 @@ class DBBitArray {
 				 typename std::enable_if<
 					 std::is_base_of<std::forward_iterator_tag, typename std::iterator_traits<Iter>::iterator_category>::value, int>::type = 0
 				 >
-		DBBitArray(Iter first, Iter last) : _storage(this->alloc(static_cast<size_type>(std::distance(first, last)))) {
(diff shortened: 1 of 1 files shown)


Ghabry:
Jenkins: Test this please <3

Ghabry (approved):
Thx. Honestly not sure why its considered uninitialized but the errors are gone :)

Ghabry:
Forwarded this change to Gemini to evaluate it: This appears to be a false positive in valgrind though I will continue merging it as it has no disadvantages.

---

Notable (I will fix this somewhere else) is that it also found a bug in `flip_all`:

> While analyzing this file, I noticed an unrelated but highly likely bug in flip_all():
code
> ``p[i] ^= p[i]; // <--- BUG``
> Any value XORed by itself is zero. Currently, flip_all() acts identically to reset_all() > (it clears all bits to 0 instead of inverting them).
> To actually flip the bits, this should be written as p[i] = ~p[i]; or p[i] ^= 0xFF;.

Though this was never noticed as this function is never used and there is no unit test for it :)

staticssleever668:
I don't think it's a false-positive, but it doesn't matter in practice either.
Memory is byte addressable. Even if we only assign and read the first n bits, we still fetch the whole byte that contains other, uninitialized bits.

